PR #31, opened 2026-05-01
Implement sandlock-oci: OCI-compliant runtime for namespace-less sandboxing

sachin2605:
# Pull Request: Implement OCI-compliant Runtime for Sandlock Sandbox

## Title
**Implement sandlock-oci: OCI-compliant runtime for namespace-less sandboxing**

## Description

This PR introduces `sandlock-oci`, a low-level OCI runtime shim that enables the `sandlock` sandbox to be used as a drop-in replacement for `runc` in `containerd`, `CRI-O`, and Kubernetes environments.

Unlike traditional runtimes, `sandlock-oci` focuses on a "no-container" approach, providing isolation via **Landlock** and **Seccomp-Notify** without relying on Linux namespaces (PID, Mount, Network, etc.). This makes it ideal for high-performance, lightweight sandboxing of host-level processes.

### Key Features
- **OCI Spec Translation**: Complete mapping of OCI `config.json` (filesystem rules, resource limits, environment) to `sandlock::Policy`.
- **Lifecycle Management**: Full implementation of OCI commands: `create`, `start`, `state`, `kill`, `delete`, and `list`.
- **Supervisor-Mediated Sandboxing**: A background supervisor process manages child lifecycle and signal synchronization (`SIGSTOP`/`SIGCONT`).
- **Persistent State**: Reliable state tracking in `/run/sandlock-oci` for integration with container engines.
- **Kubernetes Readiness**: Support for `RuntimeClass` integration, enabling `runtimeClassName: sandlock` in Pod specs.

### Implementation Details
- **`crates/sandlock-oci`**: The new OCI shim crate.
- **`src/spec.rs`**: OCI specification parsing and policy mapping logic.
- **`src/supervisor.rs`**: Background daemon logic for process orchestration.
- **`src/state.rs`**: JSON-based state persistence for OCI compliance.

### Testing
- **Integration Tests**: Comprehensive Rust test suite in `tests/integration.rs` covering the full OCI lifecycle.
- **Containerd E2E**: Integration script (`tests/containerd/test_containerd.sh`) validating manual OCI lifecycle and containerd registration.
- **Kubernetes E2E**: `kind` cluster test script (`tests/kubernetes/test_kind.sh`) for validating `RuntimeClass` and Pod deployments.

### Checklist
- [x] Integration with `sandlock-core` for Landlock/Seccomp confinement
- [x] State persistence in `/run/sandlock-oci`
- [x] End-to-end validation with `containerd` and `kind cluster`


congwang-mk:
## Code review

### Overview
Adds a new `crates/sandlock-oci` workspace member implementing OCI runtime commands (`create`/`start`/`state`/`kill`/`delete`/`list`/`check`) with a double-fork supervisor that SIGSTOPs the child until `start` is called. The architectural shape is reasonable for a namespace-less OCI shim. CI is green, but a couple of substantive concerns aren't covered by the existing CI matrix — details below.

### Blocking issues

#### 1. The translated OCI policy is discarded — sandbox does nothing
`crates/sandlock-oci/src/main.rs:484` builds a `PolicyBuilder` from the spec and immediately drops it:

```rust
let _builder = spec::spec_to_policy(&spec, &bundle)?;
```

Then at `main.rs:536`:

```rust
let policy = sandlock_core::Policy::builder()
    .build()
    .unwrap_or_else(|_| {
        sandlock_core::Policy::builder().build().expect("minimal policy")
    });
```

An *empty* policy is built and handed to the supervisor. As a result, none of the spec-derived rules are applied at runtime:
- rootfs `chroot`, `fs_read`/`fs_write`/`fs_mount`
- `process.env`, `process.cwd`
- `max_memory` / `max_processes` / `max_cpu`
- bind mounts

The `unwrap_or_else` fallback is the same call as the primary; it can't recover anything.

In addition, `confine_current_process` (`crates/sandlock-core/src/lib.rs:58`) only applies Landlock fs/IPC rules — it does not chdir, chroot, or setenv. Even if the builder were threaded through, the child still wouldn't enter the rootfs, get the spec env, or land at the spec cwd. `execvp(cmd[0], …)` resolves against the *host* `$PATH`, not the bundle's rootfs, so a containerd test that runs `busybox` actually runs the host's binary. Net effect: pods scheduled with `runtimeClassName: sandlock` are effectively unconfined, which is more dangerous than runc because users may assume sandlock-grade isolation.

CI doesn't catch this because no test creates a container, runs a process, and asserts that the rootfs/env/limits took effect. The shell scripts that would catch it aren't part of CI.

#### 2. Integration tests duplicate production code rather than test it
`tests/integration.rs:1577-1637` defines a private `sandlock_oci_test_helpers` module that re-implements stripped-down copies of `load_spec`, `spec_to_policy`, `ContainerState`, `Status`, and the setters; the assertions then run against those copies:

```rust
let spec = sandlock_oci_test_helpers::load_spec(dir.path())…
let builder = sandlock_oci_test_helpers::spec_to_policy(&spec, dir.path()).unwrap();
```

So `spec_load_and_policy_mapping` and `state_created_lifecycle` are green but they're validating the helper, not `crate::spec`/`crate::state`. The CLI-binary tests (`oci_check_exits_zero`, `oci_state_unknown_container_errors`, `oci_list_no_containers`, `oci_kill_unknown_container_errors`, `oci_delete_nonexistent_is_ok`) exercise the real binary, but only commands that don't require root and don't go through the supervisor path — i.e., none of the actual create→start→kill lifecycle.

Two ways forward: (a) add `[lib]` to `crates/sandlock-oci/Cargo.toml` re-exporting `pub mod spec; pub mod state;` and import from it directly in tests, or (b) move these checks into `#[cfg(test)] mod tests` blocks in the same `src/*.rs` files.

### Significant issues

#### 3. CI silently skips the bin's `#[cfg(test)]` tests, and they're broken
`crates/sandlock-oci/src/state.rs:215` and `:224` call `state.set_running(12345)` / `set_running(1)`, but `set_running` at `state.rs:122` takes no arguments. `cargo test -p sandlock-oci --bins` fails with two `E0061` errors. CI misses it because `.github/workflows/ci.yml:30` uses `cargo test --release --lib`, and `sandlock-oci` is a `[[bin]]` without a `[lib]` target. Worth fixing both the call sites and adding `--bins` (or a lib target) to CI so this can't drift again.

#### 4. Race on `pid_file`
`main.rs:548-559`: parent waits for the intermediate fork, sleeps 100 ms, then reads `state.pid` and writes `pid-file`. The supervisor only updates `state.pid` after binding the socket and forking the grandchild. On a busy host this can exceed 100 ms, and the pid-file gets `0`. containerd / CRI-O treat that as authoritative for tracking; a `0` will cause them to lose the container. Use a real synchronization primitive — e.g. a pipe inherited across the fork that the supervisor writes the PID to, or have the supervisor write the pid-file itself before going into `accept()`.

#### 5. `run_supervisor` accept loop is inconsistent
`supervisor.rs:1319-1361` calls `listener.set_nonblocking(false)` (blocking), but a match arm handles `WouldBlock` with a sleep. The `WouldBlock` branch is dead code as written.

#### 6. `send_command` framing is fragile
`supervisor.rs:1239-1247`: read loop breaks on `\n`, but `serde_json::to_writer` (lines 1336, 1351) writes no trailing newline. The client only exits via EOF, which only happens because each accept arm drops the stream. Works today, but small refactors (e.g. holding the stream open) would deadlock the client. Use `writeln!` on the supervisor side or a length-prefixed frame.

#### 7. `process.args` empty fallback is unsafe
`main.rs:491`: falls back to `vec!["sh".to_string()]`. OCI requires `args` non-empty for a startable container; failing fast with a clear error is better than silently running `sh` from the host.

### Smaller notes

- `spec.rs:750-754`: hard-codes `/usr`, `/lib`, `/lib64`, `/bin`, `/sbin`, `/etc`, `/proc`, `/dev` as `fs_read` and `/tmp` as `fs_write`, ignoring `root.readonly` and the `mounts` list. Push these decisions through `map_mounts` instead.
- `spec.rs:820-823`: silently skipping `proc`/`sysfs`/`devpts`/`tmpfs`/`mqueue`/`cgroup{,2}` mounts is the right call for a namespace-less runtime, but no diagnostic. A `--debug`-gated `eprintln!` would help triage.
- `state.rs:1064-1067`: `set_created(pid)` keeps `status = Created` and only updates the pid; the name suggests a status transition. `set_init_pid` is clearer.
- `state.rs`: `STATE_DIR = "/run/sandlock-oci"` is hard-coded. An env override (e.g. `SANDLOCK_OCI_STATE_DIR`) would unlock real lifecycle integration tests without root.
- `supervisor.rs`: the supervisor doesn't `chdir("/")` after the second fork, so it pins the caller's cwd.
- `state.rs`: the `Stopped` transition doesn't capture exit code or terminating signal — worth adding for CRI consumers.
- Build warnings to clean up: unused `bail` import in `spec.rs:6`, unused `FsIsolation` import in `spec.rs:8`, redundant `UnixListener` import in `supervisor.rs:9` (re-imported inside `run_supervisor`), unused `state_json` function in `state.rs:142`.
- Single commit message `"https://github.com/bytehooks/sandldd sandbox-oci crate and test cases…"` looks truncated/garbled and references an unrelated repo. Squash to a clean message before merge.
- PR head and base are both `main` — feature work on the contributor's `main` complicates rebases and reviews.
- `tests/containerd/test_containerd.sh` mutates `/etc/containerd/config.toml` and restarts `containerd`. Cleanup is `EXIT`-trapped and sed-based; ^C between `cp backup` and `restart` leaves the host containerd misconfigured.
- `tests/kubernetes/test_kind.sh` pins `kindest/node:v1.30.0` — worth an env override.

### Recommendation
Issues 1 and 2 are blocking: the runtime doesn't currently apply any spec-derived isolation, and the test suite that's keeping CI green doesn't exercise the modules in question. Once spec→policy is plumbed through to a real `chroot`/exec path inside the rootfs, and at least one end-to-end test creates a container and asserts on its confined view, the rest is small cleanup.

congwang-mk:
There are still many errors from `cargo check -p sandlock-oci`

congwang-mk:
And I added start() and create() API for OCI: https://github.com/multikernel/sandlock/pull/45
Please use them.

sachin2605:
hi @congwang-mk , please review the latest changes. I am also planning to write a k8s operator to manage the sandlock machine and publish a Helm chart too for that operator.  Please provide your thoughts on it.

crates/sandlock-oci/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+        // Redirect stdout/stderr to /dev/null to avoid polluting the caller.
+        unsafe {
+            let devnull = libc::open(
+                b"/dev/null\0".as_ptr() as *const libc::c_char,
+                libc::O_RDWR,
+            );
+            if devnull >= 0 {
+                libc::dup2(devnull, 0);
+                libc::dup2(devnull, 1);
+                libc::dup2(devnull, 2);
```
congwang-mk: Container stdio is sent to `/dev/null`. The supervisor `dup2`s fds 0/1/2 to `/dev/null` here, *before* `create_interactive` forks the child, so the container process inherits `/dev/null` for stdout/stderr.

containerd/CRI-O wire the runtime's stdio (or the `--console-socket` PTY) to the log FIFOs they create, so with this redirect all container output is discarded: `kubectl logs` and containerd log capture will be empty. Note `--console-socket` is parsed but ignored (`console_socket: _`), so there is no PTY path either.

For the "drop-in runc replacement" goal, the runtime needs to forward the container's stdout/stderr to the fds/console socket the caller provides rather than `/dev/null`. This is the most impactful OCI-compliance gap in the PR.

crates/sandlock-oci/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+    Delete {
+        /// Container identifier.
+        id: String,
+        /// Force deletion even if the container is still running.
+        #[arg(short, long)]
+        force: bool,
+    },
+
+    /// List all containers managed by sandlock-oci.
+    List,
```
congwang-mk: No `exec` subcommand. The CRI layer uses `runc exec` for `kubectl exec` and for exec-based liveness/readiness probes. Without it, those operations will fail against a `runtimeClassName: sandlock` pod. `exec` is not part of the minimal OCI command set, so this is fine to defer, but it should be tracked as a known limitation (and ideally noted in the PR description, which currently advertises full Kubernetes readiness).

crates/sandlock-oci/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+
+    match cli.command {
+        Command::Create { id, bundle, pid_file, console_socket: _ } => {
+            cmd_create(&id, &bundle, pid_file.as_deref())?;
+        }
+        Command::Start { id } => {
+            cmd_start(&id)?;
+        }
+        Command::State { id } => {
+            let state = ContainerState::load(&id)
```
congwang-mk: `state` prints the stored status verbatim and never reconciles against actual process liveness. There is an `is_alive()` helper in `state.rs` (signal-0 probe) that is not used here, so if the init process is killed out-of-band the reported status can lag reality until the supervisor's `wait()` records the exit. The OCI spec expects `state` to report the *current* status, so consider probing liveness (or transitioning to `stopped`) before printing when status is `running`.

crates/sandlock-oci/src/state.rs
```diff
@@ -0,0 +1,290 @@
...
+
+/// Return the effective state directory, respecting the env override.
+pub fn state_dir() -> String {
+    std::env::var("SANDLOCK_OCI_STATE_DIR").unwrap_or_else(|_| STATE_DIR.to_string())
+}
+
+/// OCI container status as defined by the runtime spec.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Status {
```
congwang-mk: Two minor OCI-conformance gaps around status/version:

1. The `Status` enum omits the transient `creating` state defined by the runtime spec. Low impact (it is momentary), but worth a note.
2. `ociVersion` from the bundle is stored and echoed back but never validated in `load_spec`. A bundle declaring an unsupported `ociVersion` is accepted silently. Consider rejecting versions outside the supported range so incompatible bundles fail fast rather than running with possibly-mismapped fields.

crates/sandlock-oci/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+    unsafe { libc::waitpid(pid, &mut wstatus, 0) };
+
+    // Read the child PID from the pipe — blocks until the supervisor writes it.
+    let child_pid = {
+        let mut buf = [0u8; 32];
+        let n = unsafe {
+            libc::read(read_fd, buf.as_mut_ptr() as *mut libc::c_void, buf.len())
+        };
+        unsafe { libc::close(read_fd) };
+        if n > 0 {
```
congwang-mk: Blocking: `create` reports success even when sandbox setup fails. The CLI<->supervisor pipe carries **only the PID**, with no error channel.

If `create_interactive` (or the `bind`, Landlock install, chroot, seccomp, or policy->Sandbox build) fails, `supervisor_main` returns `Err` *before* writing the PID. The daemon block does `let _ = supervisor::run_supervisor(...)` and `_exit(0)`s, so the pipe write end closes with no data. Here `read` returns `n == 0` (EOF) -> `child_pid = 0` -> the `if child_pid > 0` guard below skips the state update -> `cmd_create` returns `Ok(())` (exit 0). With `--pid-file`, the file is written with the literal `"0"`.

So a container that was never confined (or never created) looks like a successful `create` to containerd/CRI, violating the OCI requirement that `create` fail and leave no container on error; the failure only surfaces later when `start` can't reach the dead socket, and the bogus pid-file `0` is a hazard.

Fix: add an error/status variant to the pipe protocol (or a status byte) and make `cmd_create` return non-zero when the supervisor reports failure or the pipe EOFs without a PID.

crates/sandlock-oci/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+/// `sandlock-oci delete <id>`
+///
+/// Kills the container (if running) and removes the state directory.
+fn cmd_delete(id: &str, force: bool) -> Result<()> {
+    let state = match ContainerState::load(id) {
+        Ok(s) => s,
+        Err(_) => return Ok(()),
+    };
+
+    if state.status == Status::Running && !force {
```
congwang-mk: Blocking: supervisor process leak on `delete` before `start`. `create` -> `delete` (without `start`) is a legal OCI sequence.

In that state the supervisor is parked in `listener.accept().await` waiting for `Start`; it only calls `sandbox.wait()` after receiving `Start`. Here `status == Created` (not `Running`), so the `!force` guard passes, the child is `killpg`'d, and the socket + state dir are removed. The child dies (its `PR_SET_PDEATHSIG` is moot: that fires when the *supervisor* dies, not the child), but the supervisor never receives `Start`, so it blocks forever in `accept()` on a now-unlinked socket: a leaked process holding a multi-threaded tokio runtime, one per delete-before-start.

`cmd_delete` kills the container PID but never signals or reaps the supervisor. Fix: have `delete` also terminate the supervisor (a `Shutdown` command, or kill the supervisor PID), and/or give the supervisor an accept-loop timeout or child-death watch so it exits when its child is gone.

congwang-mk:
Hi @sachin2605,

Could you rebase and address the above review comments? Personally I look forward to merging this new feature.

Thanks.

Issue #43, opened 2026-05-07
RFC: Python parity for the Handler trait (Follow-up B)


dzerik:
# RFC: Python parity for the Handler trait (Follow-up B)

PR #36 landed the Rust `Handler` trait and the
`Sandbox::run_with_extra_handlers(I: IntoIterator<Item = (S, H)>)` shape.
The existing Python SDK (ctypes-based, in `python/src/sandlock/`) has no
equivalent surface — Python users can spawn a sandbox and set policy,
but cannot register user handlers.

This RFC asks for direction on five design questions before opening a
PR.  The Python SDK is currently sync ctypes over `libsandlock_ffi.so`,
which I treat as the binding constraint (no PyO3 introduction here).

## Q1: Async model

Python `async def handle()` semantics across the FFI boundary.

- **A.** Sync handler signature (`def handle(ctx) -> NotifAction`).
  Users that want async wrap themselves with
  `asyncio.run_coroutine_threadsafe(...).result()`.
  Smaller C ABI, supervisor task blocks fully on handler.
- **B.** Native async handler via completion-pipe / eventfd bridge.
  C ABI exposes a `sandlock_completion_t*` that handler signals when
  ready.  Idiomatic, but 3-4× more C surface and ctypes needs custom
  completion glue (no PyO3-asyncio equivalent).
- **C.** Handler runs in isolated Python subprocess, IPC per
  notification.  Full isolation, no GIL contention; but ~ms-per-syscall
  overhead makes high-frequency interception (VFS) impractical.

## Q2: `HandlerCtx` FFI surface

How to expose `notif`/`notif_fd`/child-memory helpers to Python:

- **A.** Fully opaque pointer + getter functions
  (`sandlock_ctx_pid(ctx)`, `sandlock_ctx_arg(ctx, idx)`,
  `sandlock_ctx_read_cstr(ctx, addr, buf, cap)`, ...).
  ABI-safe to extend; per-call FFI overhead.
- **B.** `repr(C)` struct exposed verbatim (notif_id, pid, syscall_nr,
  args[6], notif_fd inline). Direct ctypes Structure mapping.
  Zero-cost field access; freezes layout — kernel `seccomp_notif`
  changes break Python ABI.
- **C.** Hybrid: `repr(C)` notification snapshot + opaque
  `sandlock_mem_handle_t*` for child-memory access.  Notification
  data direct, memory access wrapped (sandlock controls TOCTOU
  lifetime).

[...]

PR #44, opened 2026-05-13
feat(ffi): C ABI for the Handler trait (RFC #43)

dzerik:
## Summary

Adds a C ABI surface for the `Handler` trait introduced in #36 so non-Rust callers (Python via ctypes, Go via cgo, future bindings) can register custom seccomp-notif handlers without writing Rust.

Implements the maintainer-approved options from RFC #43:

| RFC | Choice | Realization |
|-----|--------|-------------|
| Q1 (async model) | **A — sync** | C callback signature; Rust adapter offloads to `tokio::task::spawn_blocking` |
| Q2 (`HandlerCtx` surface) | **C — hybrid** | `repr(C)` snapshot (`sandlock_notif_data_t`) for the cheap copy-safe fields; opaque handle (`sandlock_mem_handle_t*`) for child-memory access |
| Q3 (`NotifAction` surface) | **C — output setters** | Caller-allocated `sandlock_action_out_t` filled via setter calls — no tagged union returned by value |
| Q4 (handler ownership) | **B — opaque box** | `sandlock_handler_t*` allocated via `sandlock_handler_new`, freed via `sandlock_handler_free` (or the supervisor on registration) |
| Q5 (exception policy) | **D — Kill default + per-handler override** | `sandlock_exception_policy_t` defaulting to `Kill` (fail-closed); applied on rc != 0, `Unset` action, or callback panic |

## Public surface added

**Rust modules** (new files under `crates/sandlock-ffi/src/`):
- `handler.rs` — opaque types, setters, accessor functions, `FfiHandler` adapter, run entry points
- `notif_repr.rs` — `sandlock_notif_data_t` stable layout

**C header** (`crates/sandlock-ffi/include/sandlock.h`):
- Types: `sandlock_notif_data_t`, `sandlock_mem_handle_t`, `sandlock_action_out_t` (+ tag enum + payload union + nested structs), `sandlock_handler_t`, `sandlock_handler_registration_t`
- Function pointer types: `sandlock_handler_fn_t`, `sandlock_handler_ud_drop_t` (both `extern "C-unwind"` for Rust handlers exposed through the C ABI)
- Functions: `sandlock_mem_read_cstr` / `_read` / `_write`, seven `sandlock_action_set_*` setters, `sandlock_handler_new` / `_free`, `sandlock_run_with_handlers` / `_interactive_with_handlers`

**sandlock-core touch** — a single `pub use` line at `crates/sandlock-core/src/lib.rs:31`:
```rust
pub use sys::structs::{SeccompData, SeccompNotif};
```
Required because the Handler trait contract already binds `SeccompNotif` through `HandlerCtx.notif`; the types must be name-reachable for any external implementer. Narrower than making `mod sys` public.

## Test plan

- [x] 39 tests pass (38 Rust integration + 1 pure-C end-to-end), 0 ignored.
- [x] `cargo build -p sandlock-ffi` clean.
- [x] `cargo clippy -p sandlock-ffi -- -A clippy::all` clean.
- [x] Coverage includes: every NotifAction variant translation, every action setter (including null-safe and payload verification), exception policies (Kill/DenyEperm/Continue), panic recovery via `catch_unwind` (verified destructively), `Unset` fallback, handler allocation/free with `ud_drop` counter assertion, `run_with_handlers` failure paths (null policy / argv / registrations / handler-in-array, with transactional cleanup), multiple-handler dispatch, real-fd `mem_read_cstr` through an intercepted `openat`, `c-unwind` ABI correctness, layout assertions via `offset_of`-style probing, k8s defense-in-depth for `child_pgid` resolution.
- [x] All claims-of-behavior tests verified by destructive sanity (the corresponding production code was temporarily mutated to confirm the test actually fails when the behavior breaks). The audit caught two "test-passes-anyway" patterns during development and fixed them: a leak-sanitizer-only assertion on `Drop` (now uses an `AtomicUsize` counter) and a self-referencing assertion on `pgid` substitution (now spawns a real child so `pid != pgid`).

## Notable design notes

**`extern "C-unwind"` for handler function pointers.** The exception-policy doc-comment advertises panic recovery via the configured policy. With plain `extern "C" fn` this was a lie — Rust 1.71+ ABI rules abort on panic across `extern "C"` boundaries. Using `extern "C-unwind"` makes the documented behavior real for Rust handlers plugged through the C ABI; pure-C callers cannot panic (C has no unwinding) so the choice is invisible to them. Stable since 1.71 (RFC 2945).

**`InjectFdSendTracked` discriminant reserved (no setter shipped).** The tracker-callback path requires a registry of pending injections that is out of scope for this PR. The enum discriminant (`SANDLOCK_ACTION_INJECT_FD_SEND_TRACKED = 5`), payload variant, and inner struct types are preserved for ABI stability; the setter is deliberately omitted. `translate_action` treats the discriminant as Unset and drains any pending srcfd before falling back to the exception policy. The setter will return in a follow-up PR that lands the tracker registry.

**Run-entry-point ownership contract.** `sandlock_run_with_handlers` always consumes the handler pointers in its registration array, regardless of return value. Early-return paths (null policy, invalid argv, invalid name, null handler in array) free the handlers via `release_registrations`, so the C caller never has to worry about a partial-consume window. Documented in `sandlock.h`.

**`child_pgid` resolution for `Kill { pgid: 0 }` substitution.** Three defense-in-depth guards in `FfiHandler::handle`:
- `notif.pid <= 0` — guards against `getpgid(0)` returning the supervisor's own pgid in nested PID namespaces (Kubernetes pod-in-pod, KubeVirt, DinD). A `killpg(supervisor_pgid)` would be a sandbox-escape vector reachable by any malicious child.
- `getpgid()` ESRCH — child exited between notification and query.
- Resolved pgid coincides with the supervisor's own — backstop for hypothetical scenarios where the child has not been moved into its own process group (see below).

In all three cases, the substitution falls back to the bare pid; the kernel rejects `killpg(pid)` if `pid` does not name a valid process group, but the supervisor survives.

## Out of scope (deferred follow-ups)

**1. `set_inject_fd_send` srcfd validation.** A C handler can today pass `srcfd = 2` (stderr) or any other supervisor-owned fd; `OwnedFd::from_raw_fd` consumes it and `close(2)` corrupts supervisor output. A simple `srcfd >= 3 && srcfd != notif_fd` check would close the most obvious foot-gun. Deferred because the right policy (deny-list vs allow-list, runtime check vs build-time) deserves your input.

**2. Unchecked `setpgid(0, 0)` calls in sandlock-core.** While auditing pgid resolution for the FFI defense-in-depth, I noticed two child-path call sites that don't check the syscall return:
- `crates/sandlock-core/src/sandbox.rs:1201` (main spawn path, right after `fork()`)
- `crates/sandlock-core/src/fork.rs:80` (COW clone child)

`context.rs:766` already checks and exits on failure. The unchecked sites are not bugs the present PR is trying to fix — sandlock correctly calls `setpgid` in all three paths — but if either failed (rare: `EPERM` if the child is already a session leader, or namespace-policy denials), the child would silently inherit the supervisor's pgid. The FFI's third guard catches this, but the underlying core would benefit from matching error handling. Happy to open a small follow-up PR adding error checks if you'd find it useful.

**3. `name` parameter via builder.** The new entry points accept `name: *const c_char` for parity with `sandlock_run`. Internally the value is applied via `Sandbox::with_name` on a clone (same pattern as `sandlock_run` in lib.rs). The core API `Sandbox::run_with_extra_handlers` does not currently accept a `name` parameter directly; happy to thread it through if you prefer that shape, but the builder-on-clone path is identical to the existing `sandlock_run`.

**4. Python bindings (PR 2).** This PR ships the C ABI only. The Python wrapper module (`sandlock.handler`, `sandlock.NotifAction`) on ctypes is the next PR per the RFC's 3-phase plan.

**5. Ergonomic helpers (PR 3).** Path-deny presets, `policy_fn`-style glue, higher-level Python idioms — third PR.

## Test environment

- Linux 6.17, x86_64
- Rust 1.71+ (`extern "C-unwind"` is stable since 1.71; `extern "C"` ABI tests cover the unchanged convention)
- python3 at `/usr/bin/python3` (two end-to-end tests spawn it; CI matrix without Python would need either the binary installed or `#[ignore]` on those two tests).

## RFC reference

Implements maintainer-approved options A1 / C2 / C3 / B4 / D5 (Kill default) from #43. Each commit explains the *why* of its scope; the audit trail across the 17 commits also shows where additional review passes caught real bugs (panic-recovery ABI, fd leaks on error paths, pgid resolution). Happy to squash on request.


congwang-mk:
Thanks for the PR.

3 high-level comments:

1) Null-handler ownership leak. When a registration array contains a NULL handler,
  run_with_handlers_inner returns without calling release_registrations, so every non-null handler leaks
  (and its ud_drop never runs) — violating the C header's "array is consumed as a whole" contract.
  One-line fix: add release_registrations(...) on the collect_registrations → None branch.

2) Sync contract is self-contradictory. The Sync impl's safety comment leans on "supervisor
  serializes per handler," but the public C-ABI docs separately tell callers ud must be "minimally
  thread-safe" — pick one. I'd prefer requiring thread-safe ud so the ABI survives a future dispatcher
  that parallelises.

3) handler.rs is a bit too large, 935 lines mixing ABI types, the FfiHandler adapter, helpers, and
   run entry points blurs the trust boundaries; splitting into handler/abi.rs + handler/adapter.rs +
  handler/run.rs would make them legible. Not blocking.

dzerik:
Thanks for the careful review — your 3 comments are addressed across 3 commits, and a follow-up audit caught 8 more bugs in the same patterns. CI is green (8/8).

## Your 3 comments

**1. Null-handler ownership leak** → `872c885` *"fix: release handlers on collect_registrations validation failure"*

You were right: `collect_registrations → None` was the one early-return path missing `release_registrations`. One-line fix as suggested. The pre-existing test was masking the bug — it manually called `sandlock_handler_free(valid)` after the failed run, which our updated C-header contract forbids ("MUST NOT call sandlock_handler_free on registered pointers"). Rewrote the test to assert the supervisor performs the release; destructively verified that removing the fix makes the test fail with counter == 0, and that restoring the old manual free causes a double-free SIGABRT.

**2. Sync contract self-contradiction** → `c2323a0` *"docs: align Sync contract — require thread-safe ud"*

Adopted your preference: the supervisor *may* dispatch concurrently across notifications, so the C caller is responsible for ensuring `ud` is thread-safe. Aligned `handler.rs` safety comment, `sandlock.h` `sandlock_handler_t` docstring, and `docs/extension-handlers.md` to the same wording. Removed the misleading "serial in practice" framing.

**3. handler.rs too large** → `fcf133a` *"refactor: split handler module into abi/adapter/run"*

Split into `handler/abi.rs` (457 lines, public ABI types + setters + accessors), `handler/adapter.rs` (270 lines, FfiHandler + translate_action + drain helpers), `handler/run.rs` (236 lines, run entry points + helpers), with `handler/mod.rs` re-exporting the public surface so external tests and downstream consumers compile unchanged. Cross-module visibility narrowed from `pub(crate)` to `pub(super)` where possible.

## Proactive audit caught 8 more

Your review prompted me to run an explicit bug-pattern audit on the rest of the branch, looking specifically for repeats of the patterns we'd already hit (doc/impl mismatches, "for show" tests, ownership gaps in error paths, defense omissions). 11 findings total — 8 fixed in this push:

### 2 critical

**`598cebb`** — supervisor-suicide vector via `killpg(0)`. The earlier "k1" fix changed `child_pgid` to 0 when `notif.pid <= 0` (nested PID namespaces, KubeVirt/k8s pod-in-pod). But `killpg(0, sig)` per POSIX kills the *caller's* process group — the supervisor. The fix-of-the-fix-of-the-fix was to introduce an `UNSAFE_PGID = i32::MIN` sentinel that propagates through `translate_action` and `exception_action`, refusing to produce a `Kill` action when no safe pgid is available (routes to exception policy → `Errno(EPERM)`). New regression test registers `SIGURG` on the test process and asserts it is *not* delivered after a lethal-pgid Kill flows through dispatch.

**`0d5b480`** — same `extern "C"` panic-abort pattern as A5, but on the high-traffic run path. The original A5 fix (`d1807c3`) made `sandlock_handler_free` use `extern "C-unwind"`. But `sandlock_run_with_handlers` / `_interactive_with_handlers` were still `extern "C"`, and `release_registrations` looped through per-element `Box::from_raw` without `catch_unwind`. A panicking user-supplied `ud_drop` aborted the process *and* skipped cleanup of remaining handlers. Switched the run entry points to `extern "C-unwind"`, wrapped each per-element drop in `catch_unwind`, accumulated the first panic, and re-raise after the loop completes — so all handlers are still freed even if one of them panics, and the caller still observes one unwinding panic at the boundary.

### 3 important

**`611aba4`** — `ud_drop` was gated on a non-null `ud` check, contradicting the C header's "invoked exactly once when the container is freed." A C caller using `ud_drop` for lifecycle logging (or any side-effect not keyed on `ud`) silently lost the cleanup. The pre-existing regression test asserted the dropper does *not* fire — locking in the bug. Removed the gate, rewrote the test to assert it fires exactly once.

(plus the test-pattern fix for the K1 supervisor-suicide test, bundled with `598cebb`)

### 5 minor

**`f312d90`** bundles:
- exact-match assertions on end-to-end test stdout (the previous `.contains("777")` was flaky on real pids — ~1% per run);
- payload-stomp checks on `set_continue` / `set_hold` (verify they're truly tag-only);
- `sandlock_mem_read_cstr` fast-path for `max_len == 1` (header promised that was sufficient for the NUL; impl was rejecting it);
- upper bounds on `argc` / `nregistrations` (both capped at 4096; without bounds, `slice::from_raw_parts(regs, usize::MAX)` against a small backing was unbounded-deref UB);
- Rust union field rename `errno → errno_value` to match the C header.

## Out of scope / deferred

Same as the original PR body, two items unchanged:

- **A4** (`srcfd >= 3` validation in `set_inject_fd_send` to prevent supervisor fd close) — still deferred pending your policy preference on deny-list vs. allow-list.
- **Unchecked `setpgid(0, 0)`** in `sandbox.rs:1201` and `fork.rs:80` — out-of-scope for this PR, happy to send a small follow-up if you'd like.

## Status

- 21 commits, +3500/-150 ish.
- 44 tests (43 Rust integration + 1 pure-C end-to-end), 0 ignored, 0 failed.
- 8/8 CI green (`Rust tests` and `Python tests` on `ubuntu-latest` + `ubuntu-24.04-arm`).
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

Happy to squash the audit-driven commits into the original feature commits before merge if you'd find that cleaner.

PR #49, opened 2026-05-16
ffi: fix sandlock_spawn failure under multi-threaded callers with restricted seccomp (#47)

### LINKED ISSUE #47: sandlock_spawn fails with ENOSYS (clone3) when called from a multi-threaded Python process (uvicorn/asyncio + Kubernetes RuntimeDefault seccomp)
## Summary

Calling `Sandbox(policy).run(...)` from a uvicorn server process returns `exit_code=-1, error=\"sandlock_spawn failed\"` every time. The identical call succeeds from a fresh single-threaded Python process in the same container.

## Context

I was setting up sandlock as the execution backend for an MCP tool server — following the recommendation in [lobehub/lobehub#12472](https://github.com/lobehub/lobehub/issues/12472) to use sandlock as a self-hosted alternative to LobeHub's cloud sandbox. Because LobeHub requires Streamable HTTP MCP transport (not SSE), I wrote a thin FastMCP wrapper around `Sandbox.run()`.

The server runs as a sidecar container in a Kubernetes k3s pod.

## Environment

- Python 3.12, sandlock 0.7.0 (pip)
- uvicorn + FastMCP (Streamable HTTP transport)
- Kubernetes k3s, kernel 6.18.18, Landlock ABI v7
- Pod seccomp: `RuntimeDefault` (Kubernetes PSS `restricted`)
- Container: UID 1000, `readOnlyRootFilesystem: true`, `allowPrivilegeEscalation: false`, `capabilities: drop ALL`

## Reproduction

Any FastMCP/uvicorn server that calls `Sandbox(policy).run()` from its request handler:

```python
@mcp.tool()
async def execute_python(code: str) -> str:
    ws = pathlib.Path("/tmp/sessions/default")
    policy = Policy(fs_readable=["/usr","/lib","/etc"], fs_writable=[str(ws)], ...)
    loop = asyncio.get_event_loop()
    return await loop.run_in_executor(None, lambda: Sandbox(policy).run(["python3", "-c", code]))
```

Result: `Result(success=False, exit_code=-1, error='sandlock_spawn failed')`

## Diagnosis

I am not a kernel developer or Python internals expert — I figured this out in collaboration with Claude Sonnet 4.6, so please correct any mistakes in the analysis.

A diagnostic endpoint injected into the running server process revealed:

```json
{
  "pid": 1,
  "active_threads": 2,
  "fork": "ok",
  "clone3": "ret=-1 errno=38 (Function not implemented)",
  "new_thread": "ok",
  "minimal_policy": {"ok": false, "error": "sandlock_spawn failed"}
}
```

Key observations:
- `fork()` works fine from the server process
- `clone3` returns `ENOSYS` — it is blocked by Kubernetes' `RuntimeDefault` seccomp profile
- Python's `threading.Thread` still works because glibc falls back from `clone3` to `clone`
- `sandlock_spawn` fails even with the most minimal policy

Reading `crates/sandlock-ffi/src/lib.rs`:

```rust
let rt = match tokio::runtime::Runtime::new() {   // = new_multi_thread()
    Ok(rt) => rt,
    Err(_) => return ptr::null_mut(),             // → "sandlock_spawn failed"
};
```

`Runtime::new()` calls `new_multi_thread()`, which spawns OS worker threads. Our hypothesis: when called from a **multi-threaded parent process** (uvicorn has 2 threads — event loop + thread pool), Tokio's worker thread spawning fails. Either `clone3` is blocked and the fallback doesn't work reliably in a multi-threaded context, or glibc's `pthread_atfork` handlers deadlock in the forked child. Python itself warns:

```
DeprecationWarning: This process (pid=1) is multi-threaded,
use of fork() may lead to deadlocks in the child.
```

The same issue exists in the current source at `lib.rs` lines ~694, ~744, ~890, ~1042, ~1224, ~1330, ~1628, ~1679, ~1710 and `handler/run.rs`.

## Workaround

Spawn a fresh single-threaded Python subprocess per sandlock call. The subprocess has no active event loop or thread pool, so Tokio's runtime creation succeeds:

```python
def _run_sandboxed_sync(cmd, ws, timeout):
    helper = r"""
import sys, json, pathlib
from sandlock import Sandbox, Policy
req = json.loads(sys.stdin.read())
# build policy, call Sandbox(policy).run(), return JSON
"""
    proc = subprocess.run(
        [sys.executable, "-c", helper],
        input=json.dumps({"cmd": cmd, "ws": str(ws), "timeout": timeout}),
        capture_output=True, text=True, timeout=timeout + 5,
    )
    return json.loads(proc.stdout)["output"]
```

This works, but adds ~50ms overhead (Python startup time) and an extra unconfined intermediary process.

## Suggested fix

Replace `Runtime::new()` with a current-thread runtime at every call site in the FFI layer:

```rust
// Before
let rt = match tokio::runtime::Runtime::new() {

// After
let rt = match tokio::runtime::Builder::new_current_thread()
    .enable_all()
    .build() {
```

A current-thread runtime runs entirely on the calling thread — no worker thread spawning, no `clone3`, no fork-safety issues. The async operations sandlock performs (waiting for child process I/O) are I/O-bound, not CPU-parallel, so there is no functional regression from dropping the multi-thread scheduler.

---

Happy to provide any additional diagnostic information or test a patched build.

congwang-mk:
## Summary

Fixes #47. `Sandbox(policy).run([...])` was failing with `sandlock_spawn failed` when called from a multi-threaded Python host (uvicorn/asyncio) on a Kubernetes pod with the `RuntimeDefault` seccomp profile.

Root cause: every FFI entry point built a fresh `tokio::runtime::Runtime::new()`, which is `new_multi_thread()` and spawns worker threads eagerly via `pthread_create`. Kubernetes' `RuntimeDefault` blocks `clone3` with `ENOSYS`, and the multi-thread builder's eager-spawn path returned `Err` before glibc's fallback could help, surfacing to the caller as a NULL handle.

Fix: switch FFI entry points to a per-thread cached `current_thread` Tokio runtime, which spawns no threads at construction. Three runtime shapes:

| Call site | Runtime | Why |
|---|---|---|
| `sandlock_run` and the rest of the one-shot entry points | thread-local `current_thread` | One-shot block_on, no tasks need to outlive the call |
| `sandlock_create` (live handle) | per-handle multi_thread, 1 worker | Supervisor must keep ticking between `start` and `wait`; one persistent worker is unavoidable here |
| `sandlock_create_for_run` (new) | per-handle `current_thread` | Python's `Sandbox.run()` is `start` then `wait` back-to-back, so suspension across the gap is fine and avoids the one worker thread `sandlock_create` would have spawned |

`Sandbox.run()` is wired to `sandlock_create_for_run`. The seccomp supervisor's blocking `SECCOMP_IOCTL_NOTIF_RECV` thread is left as-is: it spawns through `pthread_create`, which the reporter's diagnostic confirms works in their environment (`"new_thread": "ok"`).

## Test plan

- [x] `cargo build --release` clean on dev
- [x] `pytest python/tests/` 249/249 pass, including the new `TestSandlockRunCAbiMultiThreaded` and `TestSandboxRunMultiThreaded` regression tests
- [ ] Reporter retests against this branch on their k8s pod with `RuntimeDefault` seccomp (the only environment where the original failure was empirically observable)

Note: the new tests are not red-on-pristine on an unrestricted dev box because glibc's `clone3 → clone(2)` fallback masks the failure mode locally. The docstring on `TestSandlockRunCAbiMultiThreaded` documents this honestly.

### DIFF
--- a/crates/sandlock-ffi/src/handler/run.rs
+++ b/crates/sandlock-ffi/src/handler/run.rs
@@ -107,15 +107,6 @@ fn block_on_run(
     handlers: Vec<(i64, FfiHandler)>,
     interactive: bool,
 ) -> Option<Result<RunResult, SandlockError>> {
-    // Use a fresh runtime — sandlock-core already pulls in tokio with
-    // rt-multi-thread; this matches the pattern used by the existing
-    // `sandlock_run` path. A panic in an `extern "C"`-reachable path is
-    // UB, so we report runtime-build failure to the caller via `None`
-    // instead of unwrapping.
-    let rt = tokio::runtime::Builder::new_multi_thread()
-        .enable_all()
-        .build()
-        .ok()?;
     let cmd_refs: Vec<&str> = cmd.iter().map(String::as_str).collect();
     // Apply `name` via the builder method on a clone — mirrors the
     // pattern used by `sandlock_run` in lib.rs. A `None` here means
@@ -124,7 +115,11 @@ fn block_on_run(
         Some(n) => sandbox.clone().with_name(n),
         None => sandbox.clone(),
     };
-    Some(rt.block_on(async move {
+    // Drives the supervisor on the shared per-thread runtime; see
+    // `crate::runtime` for why this is `current_thread`. This path is
+    // reached from `extern "C-unwind"` entry points, so user callback
+    // panics are intentionally allowed to propagate.
+    crate::runtime::with_runtime_unwind(|rt| rt.block_on(async move {
         if interactive {
             sb.run_interactive_with_extra_handlers(&cmd_refs, handlers).await
         } else {
--- a/crates/sandlock-ffi/src/lib.rs
+++ b/crates/sandlock-ffi/src/lib.rs
@@ -14,6 +14,9 @@ use sandlock_core::{Sandbox, RunResult};
 
 pub mod handler;
 pub mod notif_repr;
+mod runtime;
+
+use runtime::{block_on_runtime, build_live_runtime, with_runtime};
 
 // ----------------------------------------------------------------
 // Opaque wrapper types
@@ -691,17 +694,13 @@ pub unsafe extern "C" fn sandlock_run(
     let args = read_argv(argv, argc);
     let arg_refs: Vec<&str> = args.iter().map(|s| s.as_str()).collect();
 
-    let rt = match tokio::runtime::Runtime::new() {
-        Ok(rt) => rt,
-        Err(_) => return ptr::null_mut(),
-    };
     let mut sb = match name {
         Some(ref n) => policy.clone().with_name(n.clone()),
         None => policy.clone(),
     };
-    match rt.block_on(sb.run(&arg_refs)) {
-        Ok(result) => Box::into_raw(Box::new(sandlock_result_t { _private: result })),
-        Err(_) => ptr::null_mut(),
+    match with_runtime(|rt| rt.block_on(sb.run(&arg_refs))) {
+        Some(Ok(result)) => Box::into_raw(Box::new(sandlock_result_t { _private: result })),
+        _ => ptr::null_mut(),
     }
 }
 
@@ -710,7 +709,10 @@ pub unsafe extern "C" fn sandlock_run(
 // ----------------------------------------------------------------
 
 /// Opaque handle for a live sandbox.
-/// Owns both the Sandbox and the tokio Runtime that drives its supervisor.
+///
+/// Owns both the Sandbox and a small Tokio runtime that drives its
+/// supervisor. Live handles need a runtime whose spawned tasks keep
+/// progressing after `sandlock_start` returns.
 #[allow(non_camel_case_types)]
 pub struct sandlock_handle_t {
     sandbox: Sandbox,
@@ -741,9 +743,9 @@ pub unsafe extern "C" fn sandlock_create(
     let args = read_argv(argv, argc);
     let arg_refs: Vec<&str> = args.iter().map(|s| s.as_str()).collect();
 
-    let rt = match tokio::runtime::Runtime::new() {
-        Ok(rt) => rt,
-        Err(_) => return ptr::null_mut(),
+    let rt = match build_live_runtime() {
+        Some(rt) => rt,
+        None => return ptr::null_mut(),
     };
 
     let mut sb = match name {
@@ -751,7 +753,7 @@ pub unsafe extern "C" fn sandlock_create(
         None => policy.clone(),
     };
 
-    if rt.block_on(sb.create(&arg_refs)).is_err() {
+    if !matches!(block_on_runtime(&rt, sb.create(&arg_refs)), Some(Ok(()))) {
         return ptr::null_mut();
     }
 
@@ -789,9 +791,9 @@ pub unsafe extern "C" fn sandlock_handle_pid(h: *const sandlock_handle_t) -> i32
 pub unsafe extern "C" fn sandlock_handle_wait(h: *mut sandlock_handle_t) -> *mut sandlock_result_t {
     if h.is_null() { return ptr::null_mut(); }
     let h = &mut *h;
-    match h.runtime.block_on(h.sandbox.wait()) {
-        Ok(result) => Box::into_raw(Box::new(sandlock_result_t { _private: result })),
-        Err(_) => ptr::null_mut(),
+    match block_on_runtime(&h.runtime, h.sandbox.wait()) {
+        Some(Ok(result)) => Box::into_raw(Box::new(sandlock_result_t { _private: result })),
+        _ => ptr::null_mut(),
     }
 }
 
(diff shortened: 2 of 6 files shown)


mrsimpson:
Verified on the target environment.

**Setup:** k3s pod, `RuntimeDefault` seccomp, Landlock ABI v7, Python 3.12, uvicorn (2 threads: event loop + thread pool).

Tested by running `Sandbox.run()` directly from within a multi-threaded Python process inside the pod:

```
threads active: 2
landlock abi: 7
success=True exit=0 stdout=b'4\n' error=None
```

The fix works. `Sandbox.run()` succeeds from the uvicorn process without any subprocess workaround. The subprocess-per-call workaround has been removed from our MCP server.

Note: the `Policy` class is gone in the dev branch — `Sandbox` is now the single dataclass for both policy and runtime. `net_allow_hosts` is now `net_allow`. We updated our code accordingly without issue.

mrsimpson:
EDIT: There are other side effects that prevent this from working e2e in my seecomp k8s. I'll continue investigation

mrsimpson:
**Follow-up: fix does not resolve the issue in our environment.**

After further testing, the dev branch does not fix the problem on our k8s pod. Here is what we found:

**What works:**
- `Sandbox.run()` called from a fresh `kubectl exec` process (any PID ≠ 1) — succeeds
- `Sandbox.run()` called from a subprocess spawned by the uvicorn server — succeeds
- Our previous subprocess-per-call workaround — still works with the new API

**What still fails:**
- `Sandbox.run()` called directly from within the uvicorn server process (PID 1)
- Both `sandlock_create_for_run` and `sandlock_run` (one-shot, thread-local runtime) return NULL
- Rust stderr is empty — no `log_build_error` output, no child error message
- This means `do_create()` returns `Err(...)` silently (swallowed by `block_on_runtime`'s match)

**Environment when failing:**
```
pid=1, active_threads=2, Seccomp: 2, Seccomp_filters: 1, NoNewPrivs: 1
```

The failure is specific to the uvicorn server process itself — not its thread count, not its seccomp state (same as exec processes), not its fd count (14 open fds, limit 1M). A subprocess spawned *by* that same process works fine.

The actual error from `do_create()` is not surfaced anywhere — `block_on_runtime` catches the `Err` variant silently and returns `None` → NULL handle. It would help to log the error string from `do_create` (e.g. to stderr) so we can see exactly which step inside `do_create` fails.

We've restored the subprocess workaround for now. Happy to test further if you have ideas on what to instrument.

mrsimpson:
**Root cause found.**

By building from a fork with `eprintln!` added to `sandlock_create_with_runtime` and `sandlock_run`, we can now see the actual error:

```
sandlock: sandlock_create_with_runtime: do_create failed:
  Runtime(Child("read notif fd from child: pipe closed before 4 bytes read"))
```

The **child process crashes** before writing the notif fd number back to the parent via `pipes.notif_w`. The parent's `read_u32_fd(pipes.notif_r)` then fails because the pipe closes.

The child's `fail!()` output goes to its stderr — which in our case is a pipe (uvicorn's stderr is `pipe:[...]`, not a tty), so it's silently swallowed before we can read it.

This happens at startup (`threads=1`, `pid=1`) — before uvicorn has even started its thread pool. So thread count is not the trigger. The child is dying in `confine_child` before it completes the seccomp/Landlock setup and calls `write_u32_fd`.

The most likely culprit: the child inherits uvicorn's `pipe:[...]` stderr fd (fd 2), and when `close_fds_above(2, &keep_fds)` runs it may close fds the child needs, or one of the earlier steps (Landlock `landlock_restrict_self`, seccomp `prctl`, or `pipe2`) is failing under the specific process state of PID 1.

Next diagnostic step: capture the child's stderr before `close_fds_above` runs, or add `eprintln!` inside `confine_child`'s `fail!()` macro so child errors reach the parent.

mrsimpson:
> **Updated summary (replaces earlier debugging comments)**

**Verdict:** PR#49 is correct and necessary, but insufficient on its own for our environment. We found a second, unrelated container-specific bug.

---

## PR#49 works as intended

The `current_thread` Tokio runtime fix correctly avoids the `clone3` seccomp issue. After fixing the second bug, `Sandbox.run()` works directly from uvicorn PID 1 with `RuntimeDefault` seccomp — no subprocess workarounds needed.

## Second bug: `getppid() == 1` in containers

- Location: `context.rs` line 788 (`confine_child`)
- Problem: `if unsafe { libc::getppid() } == 1 { fail!("parent died before confinement"); }` — in containers the entrypoint commonly runs as PID 1, so forked children legitimately have `getppid() == 1`, triggering a false positive
- Effect: child exits silently before writing to the notif pipe → parent sees *"read notif fd from child: pipe closed before 4 bytes read"*
- Fix: capture actual parent PID before fork, pass through `ChildSpawnArgs`, compare against that → https://github.com/multikernel/sandlock/pull/53

## Both fixes needed

In containerized environments (k8s, PID 1 entrypoint), both PRs are required: PR#49 for the Tokio runtime + PR#53 for the `getppid()` container check. Neither alone suffices.

congwang-mk:
Thanks for looking into this. I merged your commit first and rebased mine.

PR #53, opened 2026-05-18
fix: compare getppid() against actual parent pid, not hardcoded 1

mrsimpson:
## Summary

Fixes `Sandbox.run()` failure in containerized server processes (e.g. uvicorn running as PID 1). The child's `confine_child` checked `getppid() == 1` to detect parent death, but in containers the entrypoint process commonly runs as PID 1 itself — causing a false positive and silent child exit with `read notif fd from child: pipe closed before 4 bytes read`.

## Root cause

`context.rs` line 788: `if unsafe { libc::getppid() } == 1 { fail!("parent died before confinement"); }`

When the server process is PID 1 (container entrypoint), any forked child legitimately has `getppid() == 1`. Sandlock interprets this as "parent was reparented to init" and aborts the child.

## Fix

Capture the actual parent PID with `getpid()` before the fork and pass it through `ChildSpawnArgs`. The child compares against that value instead of hardcoded `1`.

## Testing

- Verified on k3s pod with Kubernetes `RuntimeDefault` seccomp, Landlock ABI v7, uvicorn running as PID 1
- `Sandbox.run()` now succeeds directly from the uvicorn process without any subprocess workaround
- Multiple consecutive tool calls confirmed working

## Related

- Follow-up to #47 / PR #49 (Tokio current_thread runtime fix)
- Both fixes are needed together in containerized environments

### DIFF
--- a/crates/sandlock-core/src/context.rs
+++ b/crates/sandlock-core/src/context.rs
@@ -734,6 +734,10 @@ pub(crate) struct ChildSpawnArgs<'a> {
     /// Merged into the child's BPF notif list so the kernel actually
     /// raises USER_NOTIF for them.
     pub extra_syscalls: &'a [u32],
+    /// PID of the parent process captured before fork. Used to detect
+    /// parent death in the child without assuming PID 1 is always init
+    /// (incorrect in containers where the entrypoint runs as PID 1).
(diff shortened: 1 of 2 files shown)


congwang-mk:
Oops, thanks for catching it!

PR #54, opened 2026-05-21
feat(python): ergonomic layer — read_path + 4 presets + dedicated docs page

dzerik:
Third and final PR of RFC #43. Builds on the already-merged C ABI (#44) and
Python wrapper (#46) — **no C ABI or Rust core changes**, Python and docs
only.

The minimal audit recipe now fits in five lines:

```python
from sandlock.presets import AuditPathsHandler, COMMON_PATH_SYSCALLS

audit = AuditPathsHandler(callback=lambda path, _ctx: print(f"open {path}"))
sb.run_with_handlers(cmd, [(s, audit) for s in COMMON_PATH_SYSCALLS])
```

## What's added

**`HandlerCtx.read_path(arg=None, max_len=4096) -> str | None`** — resolves a
path-bearing syscall argument. With `arg=None` (default), the path-argument
index is inferred from `ctx.syscall_nr` via a name-keyed table. Multi-path
syscalls (`renameat2`, `rename`, `linkat`, `link`, `symlinkat`, `symlink`)
and unknown syscalls raise `ValueError` — pass `arg=` explicitly in those
cases. Architecture-agnostic: the table is keyed by name and resolved at
first use via `sandlock_syscall_nr()` from the C ABI.

**`sandlock.presets`** — new module, imported explicitly (root surface stays
minimal):

- `COMMON_PATH_SYSCALLS` — the nine modern path-bearing syscalls a generic
  file-operation handler is typically registered against.
- `AuditPathsHandler(callback, max_len=4096)` — `on_exception=CONTINUE`,
  calls `callback(path, ctx)` on every intercepted syscall and returns
  `Continue`.
- `PathDenyHandler(deny: list[str], errno=EPERM, max_len=4096)` —
  `on_exception=KILL`, fnmatch patterns; `path is None → continue_()` (a
  deny-list does not claim "everything else is allowed").
- `PathAllowListHandler(allow: list[str], errno=EACCES, max_len=4096)` —
  `on_exception=KILL`, fnmatch patterns; `path is None → errno()` (an
  allow-list fails closed when the path cannot be classified).
- `LogSyscallsHandler(logger=None)` — `on_exception=CONTINUE`, default
  logger is `logging.getLogger("sandlock.audit").info`.

The deliberate asymmetry between `PathDenyHandler` and
`PathAllowListHandler` on `path is None` is documented in both class-level
docstrings and in the doc page — each class explains the security rationale.

**`docs/python-handlers.md`** — new page, single source of truth for the
Python wrapper. The Python section in `docs/extension-handlers.md` is
reduced to a one-line pointer; the Threading & safety and Ownership rules
subsections are moved verbatim. The Rust/C ABI sections in
`extension-handlers.md` are untouched.

## Out of scope (by RFC #43)

- Async handler wrappers — handlers stay synchronous per RFC Q1=A.
- pytest fixtures — pulling pytest knowledge into upstream is unjustified.
- Helpers for multi-path syscalls beyond `read_path(arg=...)` — callers
  spell the two args themselves.
- Auto-registration helpers — `COMMON_PATH_SYSCALLS` + list comprehension is
  enough.
- Helpers for non-path arguments (e.g. `read_buf`) — added if a concrete
  need appears.

## Test plan

- [x] 62 Python tests pass (unit + integration) — up from 35 baseline before
      PR 3.
- [x] End-to-end test exercises the success-criterion recipe: a real
      `python3 -c "..."` child opens a unique probe file 3 times, the
      `AuditPathsHandler` callback observes exactly 3 matching paths.
- [x] End-to-end test for `LogSyscallsHandler` against the live trampoline.
- [x] Destructive verification per load-bearing assertion — for each
      class-level claim (e.g. `on_exception=KILL` for security handlers,
      `path is None` asymmetry, default `errno`, custom `errno`), a single
      mutation of the production code makes the corresponding test fail.
- [x] Cross-arch resolution: every test that needs a syscall number uses
      `sandlock_syscall_nr()` rather than a hard-coded constant.

## Commit organization

Seven feature commits (read_path, four presets in order, e2e test, docs)
plus five polish commits addressing the deep self-review:

```
test(python): skip multi-path test when renameat2 absent on host
test(python): e2e test for LogSyscallsHandler against live trampoline
docs(python): warn read_path() raises ValueError under KILL policy
docs(python): AuditPathsHandler — warn callback can run concurrently
chore(python): clarify _reverse_path_table kernel-absent invariant
docs: dedicated python-handlers.md page; extension-handlers becomes pointer
test(python): end-to-end recipe with AuditPathsHandler + COMMON_PATH_SYSCALLS
feat(python): LogSyscallsHandler
feat(python): PathAllowListHandler
feat(python): PathDenyHandler
feat(python): sandlock.presets — AuditPathsHandler + COMMON_PATH_SYSCALLS
feat(python): HandlerCtx.read_path with name-keyed syscall table
```

Closes the RFC #43 three-PR chain.


docs/python-handlers.md
```diff
@@ -0,0 +1,279 @@
...
+- `on_exception=KILL` — security handler, fail-closed.
+- `deny` is a `list[str]` of `fnmatch` patterns; passing a single string
+  raises `TypeError` (the API is uniform).
+- If `path` matches any pattern, returns `NotifAction.errno(errno)`;
+  otherwise returns `NotifAction.continue_()`.
+- **`path is None` → `continue_()`.** A deny-list does not claim
+  "everything else is allowed", so when the path cannot be classified we
+  defer to Landlock and other handlers in the chain.
+
+### `PathAllowListHandler(allow: list[str], errno=errno.EACCES, max_len=4096)`
```
congwang-mk: PathDenyHandler vs PathAllowListHandler: one is "Deny", the other "AllowList". Pick one axis: either PathDenyHandler/PathAllowHandler or PathDenyListHandler/PathAllowListHandler. The constructor params (deny/allow) are parallel, so only the class names break the symmetry.

docs/python-handlers.md
```diff
@@ -0,0 +1,279 @@
...
+## Core types
+
+- `Handler` — base class. Subclass and override `handle(ctx) -> NotifAction`.
+  Set the class attribute `on_exception` (default `ExceptionPolicy.KILL`) to
+  choose what the supervisor does when the handler errors.
+- `HandlerCtx` — frozen dataclass with the notification fields (`id`, `pid`,
+  `flags`, `syscall_nr`, `arch`, `instruction_pointer`, `args`) plus
+  child-memory accessors.
+- `NotifAction` — frozen value-object. Construct via factory classmethods:
+  `continue_()`, `errno(value)`, `return_value_(value)`, `hold()`,
```
congwang-mk: Does return_value_() really need the trailing underscore ?

congwang-mk: errno() may conflict with the one from stdlib?

docs/python-handlers.md
dzerik: Good catch — fixed in 46adf46. Dropped "List" from the allow side so both presets are `PathDenyHandler` / `PathAllowHandler`. The `deny` / `allow` constructor params already carry the list semantics through their `list[str]` type, so the symmetric short form reads cleaner without losing meaning.

docs/python-handlers.md
dzerik: No — the underscore was only there because the factory name collided with the `return_value` dataclass field. Renamed the factory to `returns()` in 4ab093b; the field keeps its name (so the struct layout and C-ABI parity are unaffected) and the trailing underscore is gone. `NotifAction.returns(777)` reads naturally.

dzerik: No real conflict here: `NotifAction.errno(value)` is a classmethod, not a module-level name, so `import errno` (stdlib) and `NotifAction.errno(...)` coexist without shadowing — and the payload field is `errno_value`, so there's no internal collision either. The name mirrors the `ERRNO` action kind, keeping the factory set consistent (`continue_`, `errno`, `returns`, `hold`, `kill`, `inject_fd_send`). Happy to rename if you'd still prefer it, but I'd lean toward keeping it for the kind-name symmetry.

PR #61, opened 2026-05-25
Make `sandlock run` a drop-in for `docker run`

yunwei37:
## Summary

Makes `sandlock run` accept Docker's command-line shape so you can usually swap `docker run … IMAGE CMD` for `sandlock run … IMAGE CMD` unchanged — the workload runs confined by Landlock + seccomp instead of namespaces, **with no root and no daemon**. This is the "podman-style drop-in" idea: keep the muscle memory, change only the binary name.

```bash
sandlock run alpine echo hello
sandlock run -it ubuntu bash
sandlock run -e FOO=bar -w /app -v "$PWD:/app" python:3.12-slim python app.py
sandlock run --entrypoint /bin/sh alpine -c 'echo hi'
```

## What changed

**CLI shape (`sandlock-cli`)**
- The **first positional is the image** (Docker mode). A `--` terminator selects the native host-command mode (`sandlock run [flags] -- CMD`), which keeps full access to the security flags. `--image` remains as an explicit alternative.
- New Docker flags, translated onto the sandbox builder:
  - `-v/--volume HOST:CONTAINER[:ro|:rw]` → per-sandbox mount + fs read/write
  - `-w/--workdir` → working directory (cwd)
  - `-p/--publish` → net-bind + port virtualization
  - `-u/--user` → uid
  - `-e/--env`, `--env-file` → environment
  - `--hostname` → sandbox name, `--entrypoint`, `--cpus` → CPU count, `--network none|host`, `-t/--tty`, `--rm`
  - `--detach`/`--privileged`/`--cap-add`/`--cap-drop`/`--pull` are accepted for compatibility but ignored (warn).

**Image config (`sandlock-core`)**
- `image::inspect_config` now reads `ENTRYPOINT`, `CMD`, `WorkingDir`, `Env`, and `User` and applies them the way `docker run` does (image defaults sit below CLI overrides).

## Backward-incompatible flag changes

Because the first positional is now the image, the colliding short flags follow Docker. The native equivalents remain as long options:

| short | now means (Docker) | native equivalent |
|-------|--------------------|-------------------|
| `-t`  | tty                | `--timeout`       |
| `-e`  | env                | `--exec-shell`    |
| `-p`  | publish            | `--profile`       |
| `-w`  | workdir            | `--fs-write`      |

The COW storage directory moved from `--workdir` to `--fs-workdir` (since `--workdir` is now Docker's working directory).

## Limitations

The image rootfs is mounted **read-only** — rootless sandboxing has no privileged overlay mount, and the userspace branchfs needs a custom kernel FS. Use `-v` for writable host paths, or `--fs-isolation overlayfs --fs-workdir DIR` where privileges allow. Documented in the README.

## Testing

- Unit tests: docker-mode arg splitting, volume/publish/user parsing, image-config precedence.
- Docker-gated integration tests (image positional, `-e`/`-w`, image default PATH, `-v` read-only) that skip cleanly when no Docker daemon is present.
- `sandlock-cli` (unit + `cli_test` + `profile_integration`) and `sandlock-core` image tests pass.

> Note: the unrelated `sandlock-core` `test_netlink_virt::*` and `test_landlock::test_isolate_signals_blocks_parent` integration tests hang/fail in my sandboxed dev environment (they spawn privileged sandbox children); they don't touch any code in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ and **seccomp user notification** (resource limits, IP enforcement, /proc
 virtualization). No root, no cgroups, no containers.
 
 ```
-sandlock run -w /tmp -r /usr -r /lib -m 512M -- python3 untrusted.py
+sandlock run --fs-write /tmp -r /usr -r /lib -m 512M -- python3 untrusted.py
 ```
 
 ## Why Sandlock?
@@ -92,13 +92,13 @@ cargo install --path crates/sandlock-cli
 
 ```bash
 # Basic confinement
-sandlock run -r /usr -r /lib -w /tmp -- ls /tmp
+sandlock run -r /usr -r /lib --fs-write /tmp -- ls /tmp
 
 # Interactive shell
-sandlock run -i -r /usr -r /lib -r /lib64 -r /bin -r /etc -w /tmp -- /bin/sh
+sandlock run -i -r /usr -r /lib -r /lib64 -r /bin -r /etc --fs-write /tmp -- /bin/sh
 
 # Resource limits + timeout
-sandlock run -m 512M -P 20 -t 30 -- ./compute.sh
+sandlock run -m 512M -P 20 --timeout 30 -- ./compute.sh
 
 # Outbound allowlist — restrict to one host on one port
 sandlock run --net-allow api.openai.com:443 -r /usr -r /lib -r /etc -- python3 agent.py
@@ -144,7 +144,7 @@ sandlock run --net-bind 8080 -r /usr -r /lib -r /etc -- python3 server.py
 
 # Clean environment
 sandlock run --clean-env --env CC=gcc \
-  -r /usr -r /lib -w /tmp -- make
+  -r /usr -r /lib --fs-write /tmp -- make
 
 # Deterministic execution (frozen time + seeded randomness)
 sandlock run --time-start "2000-01-01T00:00:00Z" --random-seed 42 -- ./build.sh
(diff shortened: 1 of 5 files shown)


congwang-mk:
Thanks for the PR.

A quick note: it is hard to be 100% compatible with Docker CLI. Going to this direction is okay.

One small request: maybe we should use '-W' for '--fs-write', which is shorter?

crates/sandlock-cli/src/main.rs
```diff
@@ -74,11 +81,69 @@ struct RunArgs {
...
+
+    /// Network mode: `none` (default deny) or `host` (allow all egress) (Docker: --network)
+    #[arg(long = "network", visible_alias = "net", value_name = "MODE")]
+    network: Option<String>,
+
+    /// Allocate a pseudo-TTY; implies --interactive (Docker: -t/--tty)
+    #[arg(short = 't', long = "tty")]
+    tty: bool,
+
+    /// Accepted for Docker compatibility; sandboxes are always ephemeral (Docker: --rm)
```
congwang-mk, line 125: accepted but unused?

crates/sandlock-cli/src/main.rs
```diff
@@ -476,33 +688,141 @@ async fn run_command(args: RunArgs) -> Result<()> {
...
+            return Err(anyhow!("--cpus must be a positive, finite number, got: {}", n));
+        }
+        // max_cpu takes a u8; reject values whose ceiling would not fit.
+        if n > 255.0 {
+            return Err(anyhow!("--cpus must be at most 255, got: {}", n));
+        }
+        builder = builder.max_cpu(n.ceil() as u8);
+    }
+
+    // Sandbox name: --name, else --hostname, else auto-generated.
```
congwang-mk, line 790: We can just rename --name to --hostname, no reason to keep both? CLI compatibility is not a concern so far, as we are still pre-1.0.

crates/sandlock-cli/src/main.rs
```diff
@@ -110,10 +176,109 @@ struct RunArgs {
...
+    container_port
+        .parse::<u16>()
+        .map_err(|_| anyhow!("invalid --publish port in {:?}", spec))
+}
+
+/// Parse a Docker `-u/--user` value (`UID` or `UID:GID`, or a `name`), returning
+/// the numeric UID if it is numeric. Names cannot be resolved without the
+/// image's /etc/passwd, so they yield `None`.
+fn parse_user(user: &str) -> Option<u32> {
+    user.split(':').next().unwrap_or(user).parse::<u32>().ok()
```
congwang-mk, line 279: parse_user silently degrades for non-numeric users from images?

crates/sandlock-cli/src/main.rs
```diff
@@ -110,10 +176,109 @@ struct RunArgs {
...
+
+/// Parse a Docker `-p/--publish` spec `[IP:][HOST:]CONTAINER[/proto]`, returning
+/// the container port (the port the process binds inside the sandbox).
+fn parse_publish(spec: &str) -> Result<u16> {
+    let without_proto = spec.split('/').next().unwrap_or(spec);
+    let container_port = without_proto.rsplit(':').next().unwrap_or(without_proto);
+    container_port
+        .parse::<u16>()
+        .map_err(|_| anyhow!("invalid --publish port in {:?}", spec))
+}
```
congwang-mk, line 273: ' --publish' seems misleading, it only opens a bind port?

README.md
```diff
@@ -166,22 +166,69 @@ sandlock kill web.local
 sandlock run --chroot ./rootfs --fs-mount /work:/tmp/sandbox/work -- /bin/sh
 
 # COW filesystem (writes captured, committed on success)
-sandlock run --workdir /opt/project -r /usr -r /lib -- python3 task.py
+sandlock run --fs-workdir /opt/project -r /usr -r /lib -- python3 task.py
 
 # Dry-run (show what files would change, then discard)
-sandlock run --dry-run --workdir . -w . -r /usr -r /lib -r /bin -r /etc -- make build
+sandlock run --dry-run --fs-workdir . --fs-write . -r /usr -r /lib -r /bin -r /etc -- make build
```
congwang-mk, line 172: Promote short options for CLI in README ? -w and -W in this example

PR #77, opened 2026-05-30
handler: add deferred off-loop dispatch and inject_bytes content injection (core, FFI, Python)

congwang-mk:
## Summary

Adds `NotifAction::Defer` so a handler doing slow work (a network round-trip, a blocking syscall) runs off the single supervisor loop instead of stalling every other trapped syscall. The trapped child stays parked; the response is sent later, keyed by `notif.id`, while the loop keeps processing other notifications.

This branch also adds `NotifAction::inject_bytes`, a content-injection helper that pairs naturally with `Defer`: a deferred handler can fetch content off-loop and hand it to the guest as a sealed in-memory file without hand-rolling a `memfd`. See [Content injection](#content-injection-inject_bytes) below.

## Layers

**Core (sandlock-core)**
- `NotifAction::Defer(Deferred)`: a per-call return value carrying an owned `'static` future. `handle_notification` spawns a worker that drives it and responds by `notif.id`, bounded by a semaphore (`EAGAIN` on saturation).
- Refused with `EPERM` on `execve`/`execveat` and fork-creating syscalls (deferring would skip the argv-safety freeze and process creation-tracking). A nested `Defer` collapses to `EIO`. `Defer` is non-`Continue`, so it short-circuits the handler chain.
- Deferred futures need only `Send + 'static`; `NotifAction` stays `Sync` (so it can live in handler `&self` state) via a sound `unsafe impl Sync for Deferred` (the boxed future is unreachable through a shared `&Deferred`).

**FFI (sandlock-ffi)**
- `sandlock_handler_set_deferred()` plus a flag on the opaque `sandlock_handler_t` (no C ABI break). This is the deferral API for C callers, which have no `async`.

**Python**
- `async def handle` is the deferral opt-in: auto-detected at registration (`inspect.iscoroutinefunction`), its coroutine driven to completion on a worker thread (`asyncio.run`). Concurrent async handlers overlap as the GIL releases during I/O. A synchronous `handle` runs inline.

## Content injection (`inject_bytes`)

Injecting synthetic content (secrets, virtual files, fetched objects) as an `open`/`openat` result previously meant hand-rolling a `memfd` in every handler: create, write, rewind, seal, then juggle fd ownership. That dance was duplicated, and `python/examples/s3_handlers.py` shipped without sealing, leaving injected content guest-writable.

**Core**
- `content_memfd(content, seal) -> OwnedFd` builds, populates, and rewinds an in-memory file (optionally sealed read-only).
- `NotifAction::inject_bytes(content)` wraps it as a sealed, `O_CLOEXEC` `InjectFdSend` and collapses allocation failure to `Errno(EIO)`. `procfs::inject_memfd` now delegates to the primitive.

**FFI**
- `sandlock_action_set_inject_bytes(out, data, len, flags)` copies the bytes, builds the memfd, and reuses the `InjectFdSend` payload (no new action kind). `flags == 0` is the safe default (sealed + `O_CLOEXEC`); `SANDLOCK_INJECT_WRITABLE` / `SANDLOCK_INJECT_NO_CLOEXEC` opt out.

**Python**
- `NotifAction.inject_bytes(data, *, seal=True, cloexec=True)` builds the memfd in-process and routes through `inject_fd_send`; `s3_handlers.py` is refactored onto it, fixing the missing seal.

Defaults favor the dominant case (synthetic, sensitive, read-only content): sealed so the guest cannot modify or resize it, and `O_CLOEXEC` so it does not leak across the guest's `execve` into descendant processes. Both are overridable for the transparent-virtual-file case via `content_memfd` / the C flags / the Python keywords.

## Tests

- Rust: core lib 301, core integration 222 (incl. live Landlock+seccomp handler tests, and a new test where the guest reads back injected bytes for a host-nonexistent path), FFI 49 lib + C smoke + 4 handler-integration, including a pure-C `inject_bytes` check through the cdylib. All green.
- Python: 328, including discriminating e2e tests (async errno delivery, async chain short-circuit), the detection helper, and 5 new `inject_bytes` unit tests.

## Docs / example

- `docs/extension-handlers.md` and `docs/python-handlers.md` updated for both `Defer` and `inject_bytes`.
- `python/examples/s3_handlers.py` converted to `async def handle` and refactored onto `inject_bytes`; verified end to end (object read via injected memfd, `os.stat` via synthesized `struct stat`/`statx`, missing key to ENOENT).

### DIFF
--- a/crates/sandlock-core/Cargo.toml
+++ b/crates/sandlock-core/Cargo.toml
@@ -32,5 +32,5 @@ default = []
 cli = ["dep:clap"]
 
 [dev-dependencies]
-tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
+tokio = { version = "1", features = ["rt-multi-thread", "macros", "test-util"] }
 tempfile = "3"
--- a/crates/sandlock-core/src/procfs.rs
+++ b/crates/sandlock-core/src/procfs.rs
@@ -19,16 +19,14 @@
 //     contents, so the seccomp_unotify TOCTOU class doesn't apply.
 
 use std::collections::HashSet;
-use std::io::{Seek, SeekFrom, Write};
-use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+use std::os::unix::io::RawFd;
 use std::sync::Arc;
 
 use tokio::sync::Mutex;
 
-use crate::seccomp::notif::{read_child_cstr, write_child_mem, NotifAction, NotifPolicy};
+use crate::seccomp::notif::{content_memfd, read_child_cstr, write_child_mem, NotifAction, NotifPolicy};
 use crate::seccomp::state::{NetworkState, ProcessIndex};
 use crate::sys::structs::{SeccompNotif, EACCES};
-use crate::sys::syscall;
 
 // ============================================================
 // Sensitive path detection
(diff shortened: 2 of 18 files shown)


congwang-mk:
@dzerik Please take a look.

Issue #76, opened 2026-05-29
Go support: a CLI wrapper library, or just usage docs/examples?


solarhell:
## Context

I'd like to help Go users adopt sandlock and want to check the maintainers'
preference before opening a PR.

From reading the tree, the binding surface is the C ABI (`libsandlock_ffi.so`,
`crate-type = ["cdylib", ...]`), and the Python SDK is a `ctypes` layer over it.
A Go SDK would naturally be `cgo` over the same `.so`.

## The question

For the **common case** — "run an untrusted command under a policy" — Go users
are already well served by shelling out to the `sandlock` CLI via `os/exec`
(idiomatic in Go, no cgo, trivial cross-compile). A thin Go wrapper around the
CLI would add type-safe config, `--status-fd` exit/signal handling, and
`context.Context` timeouts — but it's *convenience*, not new *capability*.

So I'm unsure which is actually wanted:

1. **A small pure-Go wrapper library** (exec-based, no cgo) — ergonomic typed
   API mapping to CLI flags, captures stdout/stderr, parses `--status-fd`.
2. **Just Go usage docs + examples** showing how to drive the CLI from
   `os/exec` directly — lowest maintenance, may be all Go users need.
3. **A real cgo/FFI SDK** — only this can expose what the CLI fundamentally
   cannot (see below).

## Where the CLI is genuinely not enough

The capability that would justify an *SDK* (vs. exec'ing the CLI) is in-process
/ dynamic control that the CLI doesn't expose. Concrete example —
**command-level allow/deny** (e.g. allow `ls`/`cat`, deny `curl`):

- Via CLI you can only do **path-level** control through Landlock FS execute
  permission (`LANDLOCK_ACCESS_FS_EXECUTE`) — i.e. only grant read/exec on the
  binaries you want runnable.
- **argv-level** decisions ("allow this command, deny that one, inspect
  arguments") require intercepting `execve` in `policy_fn` (TOCTOU-safe argv
  freeze), which is **FFI-only** — there's no CLI flag for it.

Other FFI-only powers in the same bucket: `confine()` (apply Landlock to the
current process in-place), `run_with_handlers`, `checkpoint`/`fork`/`gather`.

So: an exec wrapper covers ~90% of cases cheaply; the remaining 10% is exactly
the part that needs cgo.

## What I'm proposing

I'm happy to implement (1) or (2) — whichever you prefer — and add a Go CI job
[...]

PR #83, opened 2026-06-01
Add a Go SDK (cgo bindings over libsandlock_ffi)

solarhell:
Closes the Go-support question in #76.

Per @congwang-mk's guidance there (a real cgo/FFI SDK, placed under `sandlock/go`,
following the existing FFI + Python binding), this adds a Linux-only Go SDK that
binds the sandlock C ABI (`libsandlock_ffi`) via cgo and mirrors the Python
SDK's `Sandbox` surface.

## What's included

- **Config → policy**: `Sandbox` maps every current builder field (filesystem,
  network, HTTP ACL, resource limits, syscall filtering, determinism,
  environment, COW branch handling, uid, …) to the FFI builder. `fs_isolation`
  is intentionally absent, tracking its removal in the core.
- **Execution**: `Run` (captured), `RunInteractive` (inherited stdio), `DryRun`
  (with the filesystem `Changes` list), all taking a `context.Context` (a
  deadline maps to the FFI wait timeout).
- **Lifecycle**: `Spawn` returns a `*Process` with `Pid`/`Wait`/`Pause`/
  `Resume`/`Kill`/`Ports`/`Close`.
- **In-process confinement**: `Confine` applies Landlock rules to the current
  process — something the CLI fundamentally cannot do.
- **Platform**: `LandlockABIVersion`, `MinLandlockABI`, `SyscallNr`.
- **Tests**: pure cross-platform parsing helpers in `internal/policy` with unit
  tests, plus Linux integration tests that `t.Skip` when the kernel Landlock
  ABI is below the minimum.
- **CI**: a `go` job that builds `libsandlock_ffi` and runs `go vet` + `go test`
  on the `ubuntu-latest` and `ubuntu-24.04-arm` runners.

## Verification

Built and tested in a Linux container (Landlock ABI v8): `cargo build`,
`go build ./...`, `go vet ./...`, and `go test ./...` all pass, exercising a
real sandboxed `Run`/`DryRun`. Readable paths in the tests are filtered through
`os.Stat` so the arm64 runner (no `/lib64`) stays green.

## Scope / follow-ups

This first PR covers the static policy surface plus `Confine`. Deliberately left
for follow-ups: dynamic `policy_fn` callbacks, custom seccomp handlers,
pipeline/gather, COW `fork`/`reduce`, and `checkpoint`.

One upstream note for `policy_fn`: the callback type
`sandlock_policy_fn_t` carries no `user_data`. Python works because ctypes
synthesises a distinct C function pointer per closure; Go's `//export` yields a
single fixed pointer, so it cannot route the callback to a per-`Sandbox` closure
without a `void *user_data` parameter on `sandlock_sandbox_builder_policy_fn`
(passed back into the callback). Happy to send that as a small separate PR to
unblock the Go `policy_fn` binding if you're open to it.

## Open questions

- Module path: I used `github.com/multikernel/sandlock/go` (subdir + `go/vX.Y.Z`
  tags). Happy to switch to a separate `sandlock-go` repo if you prefer.
- The default cgo link flags resolve the library at `../target/release`; let me
  know if you'd rather standardise on an installed-library path.


go/sandlock_linux.go
```diff
@@ -0,0 +1,752 @@
...
+
+// Wait blocks until the process exits, returns its captured Result, and
+// releases the handle. After Wait the Process is no longer running.
+func (p *Process) Wait() (*Result, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.h == nil {
+		return nil, ErrNotRunning
+	}
+	r := C.sandlock_handle_wait(p.h)
```
congwang-mk: Process mutex is held across the blocking native wait, so Kill can't interrupt Wait?

go/sandlock_linux.go
```diff
@@ -0,0 +1,752 @@
...
+}
+
+// Process is a live sandboxed process started by Spawn. It supports PID
+// inspection, pause/resume/kill via the process group, and Wait. A Process
+// holds at most one running command; create separate Spawns for concurrency.
+type Process struct {
+	mu  sync.Mutex
+	h   unsafe.Pointer
+	pid int
+}
```
congwang-mk, line 576: No finalizer on Process, which leaks handle + orphaned child if dropped without Close/Wait?

congwang-mk (commented):
Nice work! Just two issues below:

congwang-mk:
For the open questions, the current code is fine as it is.

solarhell:
Thanks, both are real. Fixed in 036fd89:

- **Mutex held across the wait.** `Wait` now claims the handle via a `waiting`
  flag and drops the mutex for the blocking `sandlock_handle_wait`, then
  reacquires it to free the handle. `Pause`/`Resume`/`Kill` only signal the
  process group by PID and never touch the handle, so they run while a `Wait`
  is in flight — that's what lets `Kill` interrupt it. Since the FFI handle
  isn't safe for concurrent access (`wait` borrows it `&mut`), `Ports` reports
  empty and `Close` defers the actual free to the in-flight `Wait` rather than
  aliasing the handle. Added `TestProcessKillInterruptsWait` to cover it.

- **No finalizer.** `Spawn` now installs a `runtime.SetFinalizer` that kills
  the group and frees the handle if a `Process` is dropped without
  `Wait`/`Close`; both clear the finalizer once they've cleaned up.

Verified green in a Linux container (Landlock ABI v8): cargo build, go
vet/build/test, gofmt.

go/sandlock_linux.go
```diff
@@ -0,0 +1,752 @@
...
+const uint8_t* sandlock_dry_run_result_stderr_bytes(const void* r, size_t* len);
+size_t         sandlock_dry_run_result_changes_len(const void* r);
+char           sandlock_dry_run_result_change_kind(const void* r, size_t i);
+char*          sandlock_dry_run_result_change_path(const void* r, size_t i);
+void           sandlock_dry_run_result_free(void* r);
+
+int     sandlock_landlock_abi_version(void);
+int     sandlock_min_landlock_abi(void);
+int64_t sandlock_syscall_nr(const char* name);
+*/
```
congwang-mk, line 13: I fixed the sandlock.h generation (PR #87), you can just use the existing sandlock.h:
```c
  /*
  #cgo CFLAGS: -I${SRCDIR}/../crates/sandlock-ffi/include
  #include "sandlock.h"
  */
  import "C"
```

solarhell:
Done in e3c7e58 — merged latest `main` and switched the cgo preamble to the
generated header:

```c
#cgo CFLAGS: -I${SRCDIR}/../crates/sandlock-ffi/include
#include "sandlock.h"
```

That drops the ~110 hand-written prototypes, so the bindings can't drift from
the FFI anymore. The generated types also let me replace the `void*` /
`unsigned char` placeholders with the named opaque structs, `C.bool` for the
bool setters/getters, and `uintptr_t` for the byte-length and change-index
params. Public Go API is unchanged.

Verified green against the merged `main` in a Linux container (Landlock ABI
v8): cargo build, go vet/build/test, gofmt. Thanks for fixing the header
generation!

PR #98, opened 2026-06-10
OCI compliant runtime for sandlock

sachin2605:
check https://github.com/multikernel/byte/issues/5

PR #100, opened 2026-06-10
oci: review follow-ups (cleanup, root.readonly, tmpfs mounts, lifecycle fixes)

congwang-mk:
Follow-ups to #98 (the merged OCI runtime), covering the remaining review
comments from #31 plus a lifecycle bug found while auditing the
create/start API usage.

## Changes

- Remove the dead `--debug` CLI flag (parsed but never read).
- Remove the unused `ExitInfo::from_status` helper (only its own tests
  referenced it); clears the last build warnings.
- Honor `root.readonly`: grant the whole chroot read-only vs. read-write per
  the flag, replacing the hardcoded `/usr,/lib,...` + always-writable `/tmp`
  heuristic. Images with files outside the old fixed list now work.
- Translate OCI mounts by intent in `map_mounts`:
  - writable scratch tmpfs (`/tmp`, `/run`, `/dev/shm`) -> host backing dir
    under the container state dir, bind-mounted read-write (works on a
    read-only root, isolated from the rootfs, removed with the state dir on
    `delete`);
  - `/proc`, `/sys`, and a tmpfs `/dev` -> read passthrough so sandlock's
    procfs virtualization and `/dev` interception serve them (an empty backing
    dir would shadow `/dev/null`, `/proc/*`, etc.);
  - `devpts`/`mqueue`/`cgroup` -> skipped (no safe namespace-less equivalent).
- chdir the supervisor daemon to `/` so it does not pin the caller's cwd.
- Fix a supervisor hang: a failed `start()` no longer falls through to
  `wait()` (which would block forever on the still-parked child and leak the
  supervisor); it now exits via the `Drop`-based cleanup path.
- Doc fix: the `set_created` comment no longer references the removed SIGSTOP
  parking mechanism.

## Testing

- Warning-free build.
- All sandlock-oci tests pass (27 lib + 30 lib+bin + 8 integration).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,6 +64,7 @@ jobs:
         run: |
           mkdir -p dist
           cp target/${{ matrix.target }}/release/sandlock dist/ 2>/dev/null || true
+          cp target/${{ matrix.target }}/release/sandlock-oci dist/ 2>/dev/null || true
           cp target/${{ matrix.target }}/release/libsandlock_ffi.so dist/ 2>/dev/null || true
           tar -czf sandlock-${{ matrix.target }}.tar.gz -C dist .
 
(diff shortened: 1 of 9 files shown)


congwang-mk:
@sachin2605 Please take a look. Thanks.

PR #101, opened 2026-06-10
oci: honor process.user via gid-capable run-as (single-id userns)

congwang-mk:
Wires OCI `process.user` (uid/gid) through to the sandbox, and along the way
makes sandlock-core's run-as identity gid-capable (the old `--uid` forced
`gid = uid`, which is wrong for OCI where uid and gid are independent).

Branched off `main`, so it does not include the PR #100 follow-ups; both
touch `crates/sandlock-oci/src/policy.rs`, so a small merge conflict is
expected when the two land.

## sandlock-core

- Replace the single `--uid N` (which mapped one id and forced `gid = uid`)
  with a gid-capable run-as identity:
  - `RunAs { uid, gid }`, `builder.user(uid, gid)`, and the CLI flag
    `--user UID:GID` (both ids required, no implicit default).
  - `confine_child` maps `gid` independently (fixes the old `gid = uid` bug).
- Skip the user namespace entirely when the requested uid/gid already match
  the runtime's: there's no point unsharing to map an identity already held,
  and it avoids imposing an unprivileged-userns requirement on callers that
  don't need a remap.
- Mechanism unchanged in spirit: a single-entry, unprivileged user namespace
  (`unshare(CLONE_NEWUSER)` + `uid_map`/`gid_map`), no host privilege.
- Propagated through the FFI (`sandlock_sandbox_builder_user(b, uid, gid)`,
  header regenerated with cbindgen) and the Go (`GID`) and Python (`gid`) SDKs;
  the profile (`[program]`) now requires `uid` and `gid` together; docs updated.

## sandlock-oci

- `OciPolicy` carries `run_user` from `process.user`; `to_sandbox` calls
  `builder.user(uid, gid)` unconditionally and lets core decide whether a
  user namespace is actually needed.

## Scope and limitations

- Single-id mapping only: `additional_gids` and `umask` from `process.user`
  are not applied (an unprivileged user namespace maps one id and must deny
  `setgroups`). Range mapping (`newuidmap`/subuid) is intentionally out of
  scope; it is not needed for the common case (most images run as root, and
  in rootless containerd the runtime is already ns-root in rootlesskit's range
  userns, so the requested `0:0` matches and no extra userns is created).

## Testing

- `cargo build --workspace`; core lib 398 tests pass; sandlock-oci lib 27 pass.
- New tests: `RunAs` parsing requires `UID:GID`; the userns is skipped when the
  identity matches the runtime (exercised without unprivileged userns); OCI
  `process.user` populates `run_user` and reaches `builder.user`.
- Not verified end-to-end under a real (rootless) containerd; the matching/skip
  path is covered, the differing-uid userns path is correct by construction but
  unverified in this environment (unprivileged `uid_map` writes are blocked here).

### DIFF
--- a/crates/sandlock-cli/src/main.rs
+++ b/crates/sandlock-cli/src/main.rs
@@ -394,7 +394,7 @@ async fn run_command(args: RunArgs) -> Result<i32> {
         // Network virtualization
         b = b.port_remap(base.port_remap);
         // Process identity
-        if let Some(uid) = base.uid { b = b.uid(uid); }
+        if let Some(user) = base.user { b = b.user(user.uid, user.gid); }
         // Hardware constraints
         if let Some(ref devs) = base.gpu_devices { b = b.gpu_devices(devs.clone()); }
         if let Some(ref cores) = base.cpu_cores { b = b.cpu_cores(cores.clone()); }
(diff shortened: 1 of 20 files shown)


congwang-mk:
@sachin2605 Please take a look. Thanks.

PR #107, opened 2026-06-22
ci: add ubuntu-22.04 low-ABI job + end-to-end degrade test

dzerik:
Follow-up to the CI split-out agreed in #71 (the `ci.yml` review thread).

## Motivation

The CI matrix only runs v6-capable runners (`ubuntu-latest`, `ubuntu-24.04-arm`), so the Protection opt-out path added in #71 — which lets a sandbox confine on kernels below Landlock ABI 6 by degrading the scopes the host lacks — has no automated coverage on a real low-ABI kernel. The synthetic `test_protection` resolution tests cover the mechanics, but nothing exercised a genuine `build()` + `run()` below v6.

## What this adds

- A dedicated `rust-low-abi` job on `ubuntu-22.04`. That image runs a 6.8 Azure kernel — **Landlock ABI v4** — below the project's `MIN_ABI = 6`. The default `strict_all()` suite (`test_sandbox`, `test_network`, …) hard-requires v6 and would panic there, so the job runs only the host-ABI-independent resolution tests plus a new end-to-end degrade test, and reports the runner's ABI via `sandlock check` (so a future runner-image kernel bump that changes it is visible at a glance).
- `degraded_policy_confines_and_runs_below_v6`: builds a fully-degradable policy, asserts every scope the host cannot provide resolves to `Degraded` (not `Unavailable`, which is what makes `confine` refuse), and runs a genuinely confined child. This is the test that gives a sub-v6 runner its purpose.
- A `workflow_dispatch` trigger for manual runs.

## Verification

Ran on real kernels across the low-ABI spectrum:

- ABI v1 — Ubuntu 22.04, kernel 5.15
- ABI v4 — the GHA `ubuntu-22.04` runner itself (confirmed by the job's own `sandlock check` output)
- ABI v5 — Rocky 9.6, kernel 5.14
- ABI v8 — dev host

`test_protection` is 22/22 on every host; the degrade assertion fires on each sub-v6 kernel and the confined child runs.

## Note on the `ubuntu-latest` check

`Rust tests (ubuntu-latest)` currently fails on `main` independently of this change: `test_restore::test_restore_real_program_resumes` fails on x86 (it passes on `ubuntu-24.04-arm`), introduced with the checkpoint/restore merge (#105). This PR doesn't touch that path, and the new `rust-low-abi` job is green. Happy to file that separately if it isn't already tracked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  workflow_dispatch:
 
 jobs:
   rust:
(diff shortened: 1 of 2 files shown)


congwang-mk:
@dzerik Could you fix the CI failure?

PR #109, opened 2026-06-22
sandlock-oci: implement OCI exec (supervisor-hosted, non-TTY)

congwang-mk:
## Summary

Implements the OCI `exec` command in `sandlock-oci` so containerd/CRI-O (and thus `kubectl exec` and exec-based liveness/readiness probes) can run an additional, fully-confined process inside a running sandlock container. This was the crate's last known limitation.

## Design (supervisor-hosted, non-TTY)

sandlock has no PID namespace to `setns` into, so `exec` cannot attach to the container's main process. Instead it launches a new process under the same policy, spawned by the container's long-lived supervisor daemon so the daemon owns every process in the container (shared session, process group, and seccomp supervision).

Because the daemon, not the CLI, is the parent, the CLI's stdin/stdout/stderr (which the container shim wired to the exec stream) are passed to the daemon over the control socket via `SCM_RIGHTS`. The daemon spawns the child with those fds via `Sandbox::create_with_io`, folds it into the container's process group with `setpgid` (so `kill --all` / `delete --force` reap it), and relays a two-phase `Pid` then `Exit` reply. Attached exec exits with the child's status; `--detach` returns after `Pid`.

Exec'd processes are confined identically to the container's main process: virtual chroot (notify path rewriting), seccomp filter, network ACL, and resource limits, all rootless.

## Scope

- Supports `--process <spec.json>` and inline args, with `--env` / `--cwd` overrides merged on top.
- `--detach`, `--pid-file`.
- `-t` / `--console-socket` are accepted for runc compatibility but ignored. PTY support is deferred (the confinement model already permits terminal ioctls, so it is a clean follow-up); exec runs non-TTY only for now.

## Implementation (6 commits)

1. `SCM_RIGHTS` fd-passing helpers (`fdpass.rs`) with a roundtrip unit test that proves a passed fd is the same open file.
2. `SupervisorCmd::Exec` and `SupervisorReply::Exit` protocol additions.
3. Exec CLI struct and process-spec / env / cwd resolution.
4. Daemon `handle_exec` (clone policy, `create_with_io`, `setpgid`, two-phase reply, wait in a spawned task so the serve loop never blocks) plus the CLI socket client.
5. Landlock-gated end-to-end test: create+start a long-lived container, exec a command into it, assert the CLI exits 0 and the exec'd process ran inside the container chroot.
6. Docs: record exec in the lifecycle and note the non-TTY limitation.

## Testing

- Full `sandlock-oci` suite green: 42 lib + 48 bin + 13 integration on a Landlock v8 / x86_64 host.
- The e2e exec test passed 12 consecutive runs. It uses the prebuilt `rootfs-helper` (portable C, no inline freestanding program, no arch guard).

### DIFF
--- a/crates/sandlock-oci/src/fdpass.rs
+++ b/crates/sandlock-oci/src/fdpass.rs
@@ -1,0 +1,165 @@
+//! SCM_RIGHTS file-descriptor passing over the supervisor control socket.
+//!
+//! `exec` is the only control command that carries open fds (the CLI's
+//! stdin/stdout/stderr, which the container shim wired to the exec stream).
+//! The daemon must receive those fds with the SAME `recvmsg` that reads the
+//! command bytes, because ancillary data binds to specific bytes. All other
+//! commands send an empty ancillary payload, so the receive path is uniform.
+
+use std::io;
+use std::os::unix::io::{FromRawFd, OwnedFd, RawFd};
+
+/// Send `data` plus `fds` (as SCM_RIGHTS) over a blocking unix socket.
+pub fn send_with_fds(
+    stream: &std::os::unix::net::UnixStream,
+    data: &[u8],
+    fds: &[RawFd],
+) -> io::Result<()> {
+    use std::os::unix::io::AsRawFd;
+
+    let mut iov = libc::iovec {
+        iov_base: data.as_ptr() as *mut libc::c_void,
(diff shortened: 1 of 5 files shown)


congwang-mk:
This design is flawed. Close this one for a new design.

Issue #108, opened 2026-06-22
Go SDK: Protection opt-out parity (AllowDegraded / Disable)


dzerik:
@solarhell — heads-up on a small parity gap in the Go SDK, and an offer.

PR #71 added the Protection opt-out (`allow_degraded` / `disable` per Landlock protection) so a sandbox can run below the default ABI v6 floor by degrading or disabling the v6-only scopes. It landed in Rust, Python, and the C ABI, but the Go SDK didn't pick it up: `Sandbox` has no `AllowDegraded` / `Disable`, and `go/sandbox.go` + the README still state a hard "requires Linux 6.12+ (Landlock v6)" — no longer the real floor once you opt out.

The C symbols are already in the generated header (`sandlock_sandbox_builder_allow_degraded`, `sandlock_sandbox_builder_disable`, `sandlock_protection_min_abi`), so the binding is mechanical: a `Protection` type mirroring the discriminants, `AllowDegraded` / `Disable []Protection` fields wired into `buildPolicy`, a `ProtectionMinABI` helper, and a doc tweak.

I wired this for our own use and it's working — `go vet` clean, and I validated the end-to-end degrade path on real low-ABI kernels (a fully-degradable policy builds and confines on Landlock ABI v1 and v5, where the default strict policy fails to build). Happy to open it as a PR for your review, or hand you the patch / leave it entirely to you — the Go SDK is your call. Let me know which you'd prefer.

PR #117, opened 2026-06-28
feat(go): Protection opt-out parity — AllowDegraded / Disable

dzerik:
Brings the Go SDK to parity with the Protection opt-out added in #71 and
already exposed by the Python SDK (#46/#54), closing the gap tracked in #108.

## What

By default every Landlock protection is enforced strictly, so the build
fails on a kernel below the protection's floor (the highest being ABI v6).
This lets a Go sandbox run below v6 by degrading or disabling the v6-only
protections, mirroring the Python `allow_degraded` / `disable` surface:

- `Protection` type with the six C-ABI discriminants (`ProtectionFSRefer` …
  `ProtectionAbstractUnixSocketScope`).
- `Sandbox.AllowDegraded` / `Sandbox.Disable` fields, wired through the
  builder in `buildPolicy`. `Disable` is applied last, so it wins when a
  protection appears in both lists (documented on the fields).
- `ProtectionMinABI` exposes `sandlock_protection_min_abi` for capability
  checks.

No new C ABI — this binds the existing `sandlock_sandbox_builder_allow_degraded`,
`sandlock_sandbox_builder_disable`, and `sandlock_protection_min_abi`.

## Tests

- `TestProtectionMinABI` pins the discriminants against the live FFI: a
  wrong constant value would resolve to a different protection (or unknown)
  and fail the expected `min_abi` (2/3/4/5/6/6).
- `TestRunDegradedBelowV6` confines and runs a fully-degradable policy end
  to end. Below v6 it first asserts the default strict policy fails to build,
  so the degraded run's success is attributable to the opt-out rather than
  to the host satisfying the protections anyway. Self-skips without Landlock.

`go vet -tags sandlock_repo ./...` and `go test -tags sandlock_repo ./...`
pass; `gofmt` clean.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -5,8 +5,9 @@ lightweight Linux process sandbox built on Landlock, seccomp-bpf, and seccomp
 user notification. No root, no Docker, no namespaces.
 
 The bindings bind the sandlock C ABI (`libsandlock_ffi`) via cgo, mirroring the
-Python SDK's `Sandbox` surface. **Linux only**; the runtime requires Linux
-6.12+ (Landlock ABI v6).
+Python SDK's `Sandbox` surface. **Linux only**; by default the runtime requires
+Linux 6.12+ (Landlock ABI v6), but the `AllowDegraded` / `Disable` fields let a
+sandbox run on older kernels by degrading or disabling the v6-only protections.
 
 ```go
 import sandlock "github.com/multikernel/sandlock/go"
--- a/go/sandbox.go
+++ b/go/sandbox.go
@@ -3,8 +3,10 @@
 // notification. It binds the sandlock C ABI (libsandlock_ffi) via cgo and
 // mirrors the Python SDK's Sandbox surface.
 //
-// The bindings are Linux-only. The runtime requires Linux 6.12+ (Landlock
-// ABI v6); see the project README for the full kernel feature matrix.
+// The bindings are Linux-only. By default the runtime requires Linux 6.12+
+// (Landlock ABI v6); use the AllowDegraded or Disable fields to run on older
+// kernels by degrading or disabling the v6-only protections. See the project
+// README for the full kernel feature matrix.
 //
 // # Building
 //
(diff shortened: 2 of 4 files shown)


congwang-mk:
**Doc-comment placement: `Protection` block detaches `Sandbox`'s documentation**

In `go/sandbox.go`, the new `Protection` type + const block is inserted between the `Sandbox` doc comment and `type Sandbox struct`:

```go
// A Sandbox value carries no runtime state: Run, RunInteractive, and DryRun
// build a fresh native policy on each call, so a single Sandbox may be reused
// and shared across goroutines. Use Spawn for explicit process lifecycle
// control, which returns an independent *Process handle.
// Protection identifies a single Landlock protection whose enforcement
// posture can be opted out of via the Sandbox AllowDegraded / Disable fields.
// The values match the sandlock C ABI discriminants.
type Protection uint32
```

By Go's doc-comment rules a comment must be adjacent to the declaration it documents. As written, the `Sandbox` paragraph and the `Protection` paragraph merge into one comment attached to `type Protection uint32`:

- `go doc .../go Protection` shows the unrelated `Sandbox` prose, and
- `go doc .../go Sandbox` (the package's primary type) shows no doc comment at all.

`gofmt`/`go vet` don't flag this, so it passes the PR's stated checks while still regressing the public docs.

**Fix:** move the `Protection` type + const block to *before* the `Sandbox` doc comment (or *after* the `Sandbox` struct). Placing it between them leaves `Sandbox` undocumented regardless of blank lines, since the whole block sits between the comment and the struct.

The rest of the PR checks out: the discriminant values (0-5) and ABI floors (2/3/4/5/6/6) match `sandlock_core::Protection` and the FFI exactly, and the `AllowDegraded`-then-`Disable` ordering in `buildPolicy` correctly gives `Disable` last-wins precedence.

dzerik:
Thanks, good catch. Moved the `Protection` type + const block above the `Sandbox` doc comment so each comment is adjacent to its own declaration — `Sandbox` is documented again and `Protection` keeps its own doc. Rebased onto current `main`. (393d673)

Issue #67, opened 2026-05-26
RFC: Add streaming-stdio process API (`popen` / `Process`)


congwang-mk:
## Goal

Add a `popen()`-style entry point on `Sandbox` that returns a live process handle exposing streaming stdin / stdout / stderr, so callers can drive bidirectional stdio protocols (MCP, LSP, REPLs, JSON-RPC over stdio) without losing the kernel-enforced jail.

## Motivation

Today's `Sandbox` API is capture-mode only:

- `Sandbox.run(cmd)` blocks and returns a `Result` with buffered `stdout` / `stderr`.
- `Sandbox.spawn(cmd) + wait()` exposes a live pid but no stream access; you still only get final buffered output.

That works for batch tools. It does not work for any subprocess whose protocol is request / response streamed over stdio while the process is alive. Concrete cases this blocks today:

- Running an untrusted MCP server inside a jail and connecting an MCP client to its stdio.
- Driving an LSP server in a jail.
- Hosting a REPL or interactive process whose output must be read as it arrives.
- Anything JSON-RPC over stdio.

The available workarounds are unattractive: either skip the Python API and shell out to `sandlock run` so a higher-level SDK can manage stdio, or build private stdio plumbing on top of the FFI. Neither composes well; both push complexity onto every caller.

## Current state

- `Sandbox.run` / `spawn` / `wait` (Python): capture-mode, see `python/src/sandlock/sandbox.py:463, 744, 757`.
- `sandlock_handle_*` (FFI): no pipe-fd accessors; `sandlock_handle_wait*` always buffers stdout / stderr into the returned result.
- `crates/sandlock-core/src/sandbox.rs`: launch path always wires stdio for capture.

The Python `Sandbox` object also conflates two roles: it is a policy dataclass and a single-process controller (handle is stashed in `_handle`, so one `Sandbox` instance can only run one process at a time). That wart blocks straightforward "launch many processes from one policy" patterns.

### Python

```python
class Sandbox:
    async def popen(
        self,
        cmd: Sequence[str],
        *,
        stdin:  bool = True,
        stdout: bool = True,
        stderr: bool = True,
    ) -> "Process": ...

class Process:
    pid: int
    stdin:  anyio.abc.ByteSendStream    | None
    stdout: anyio.abc.ByteReceiveStream | None
    stderr: anyio.abc.ByteReceiveStream | None

    async def wait(self) -> int: ...      # exit code only
    def terminate(self) -> None: ...      # SIGTERM
[...]

PR #118, opened 2026-06-28
feat(core): streaming-stdio popen → Process with per-stream StdioMode

dzerik:
First of a small PR series implementing the streaming-stdio process API
(RFC #67): **core** here, then the FFI surface, then the Python wrapper.

## What

`Sandbox::run` is capture-only — it buffers stdout/stderr into a `RunResult`
after the process exits. That cannot drive a process whose protocol is
request/response over stdio while it is alive (an MCP or LSP server, a REPL,
JSON-RPC). This adds `Sandbox::popen`, which generalizes the internal
`do_create` capture/inherit boolean into a per-stream
`StdioMode { Inherit, Piped, Null }` and returns a live `Process` whose
`Piped` stream ends the caller owns via `take_stdin`/`take_stdout`/`take_stderr`.

- `do_create` is now a thin wrapper over `do_create_stdio`, so
  `run`/`spawn`/`create` and their interactive variants keep **byte-identical**
  capture/inherit behavior.
- `StdioMode` is `#[repr(u32)]` with pinned discriminants so the FFI/Python
  layers can bind a stable C enum.
- `Null` wires the stream to `/dev/null` and **fails closed**: if `/dev/null`
  cannot be opened (or `dup2` fails) the fd is closed, never left inherited
  from the supervisor.
- A piped stdin the caller never takes is closed by `wait()` so a child
  reading stdin still reaches EOF instead of hanging.
- `popen` does not poll for `execve` (a streaming reader blocks until bytes
  arrive; the poll would spuriously time out on a fast-exiting child).
- `Process` *borrows* the `Sandbox` rather than owning the process (unlike
  `std::process::Child`); the child is killed and reaped by `Sandbox::drop`
  or `Process::kill` (whole process group), and the type is `#[must_use]`.
  Folding the legacy `spawn`/`wait` into `Process` is deferred to keep this PR
  small.

## Tests

Run on a multi-threaded runtime — a blocking pipe read must not starve the
seccomp-notify supervisor (documented on `popen`):

- stdin+stdout roundtrip through `cat`;
- stdout-only `echo`, stderr-only piped (guards the fd-2 vs fd-1 arm);
- `Null` proven to be `/dev/null` (the child stats its own fd 1 — fails if
  `Null` fell back to inherit or a pipe);
- an untaken piped stdin still reaches EOF at `wait`;
- dropping the owner kills the whole process group, reaching a forked
  grandchild (reaped via init);
- a regression that capture-mode `run` still buffers stdout.

`cargo test -p sandlock-core` is green (lib + integration). Pre-existing
`test_cow` / `test_handlers` failures on my host reproduce on `main` and are
unrelated.

Refs #67

### DIFF
--- a/crates/sandlock-core/src/sandbox.rs
+++ b/crates/sandlock-core/src/sandbox.rs
@@ -201,6 +201,47 @@ pub enum BranchAction {
 // Runtime — private heap-allocated state, present only while running
 // ============================================================
 
+/// How one of a child's standard streams (stdin/stdout/stderr) is wired.
+///
+/// `Inherit` shares the supervisor's own fd (the child writes to the same
+/// terminal/file the parent has). `Piped` creates a pipe whose caller-side end
+/// is handed out via [`Process`] so the caller can stream to/from the live
+/// process. `Null` connects the stream to `/dev/null`.
+///
+/// The discriminants are a stable contract: the FFI/Python bindings pass them
+/// as a `u32`, so they are pinned with `#[repr(u32)]`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum StdioMode {
+    /// Inherit the supervisor's corresponding fd.
+    Inherit = 0,
+    /// Connect to a pipe; the caller owns the other end (see [`Process`]).
+    Piped = 1,
+    /// Connect to `/dev/null`.
+    Null = 2,
+}
+
+/// Per-stream stdio wiring for a child process.
+#[derive(Debug, Clone, Copy)]
+struct StdioSpec {
+    stdin: StdioMode,
+    stdout: StdioMode,
+    stderr: StdioMode,
+}
+
+impl StdioSpec {
+    /// Capture mode used by `run`/`spawn`: stdin inherited, stdout/stderr piped
+    /// and drained into the `RunResult` by `wait`.
+    fn capture() -> Self {
+        StdioSpec { stdin: StdioMode::Inherit, stdout: StdioMode::Piped, stderr: StdioMode::Piped }
+    }
+
+    /// Interactive mode: every stream inherits the supervisor's fd.
+    fn inherit() -> Self {
+        StdioSpec { stdin: StdioMode::Inherit, stdout: StdioMode::Inherit, stderr: StdioMode::Inherit }
+    }
+}
+
 /// Private runtime state.  Only allocated after `start()` / `run()` is
 /// called; `None` for config-only `Sandbox` instances.
 struct Runtime {
@@ -213,6 +254,9 @@ struct Runtime {
     loadavg_handle: Option<JoinHandle<()>>,
     _stdout_read: Option<std::os::fd::OwnedFd>,
     _stderr_read: Option<std::os::fd::OwnedFd>,
+    // Parent-held write end of a piped stdin (popen). The caller takes it via
+    // `Process::take_stdin`; closing it signals EOF to the child.
+    _stdin_write: Option<std::os::fd::OwnedFd>,
     seccomp_cow: Option<crate::cow::seccomp::SeccompCowBranch>,
     supervisor_resource: Option<Arc<tokio::sync::Mutex<crate::seccomp::state::ResourceState>>>,
     supervisor_cow: Option<Arc<tokio::sync::Mutex<crate::seccomp::state::CowState>>>,
@@ -660,6 +704,11 @@ impl Sandbox {
             });
         }
 
+        // Deliver EOF to a piped stdin the caller never took: otherwise a child
+        // that reads stdin (e.g. `cat`) blocks forever and this wait never
+        // returns. A taken stdin is already None here (the caller owns it).
+        drop(self.rt_mut()._stdin_write.take());
+
         // Wait for the top-level child to exit. Prefer the child's pidfd via
         // `AsyncFd`: pidfd readiness fires only on *exit*, so — unlike a
         // `waitpid` loop — it never consumes the child's ptrace-stops, which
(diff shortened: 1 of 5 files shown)


crates/sandlock-core/src/sandbox.rs
```diff
@@ -1483,14 +1568,70 @@ impl Sandbox {
...
+                StdioMode::Piped => {
+                    if let Some((ref r, _)) = stdin_p {
+                        unsafe { libc::dup2(r.as_raw_fd(), 0) };
+                    }
+                }
+                StdioMode::Null => {
+                    // Fail closed: if /dev/null can't be opened (or dup2 fails),
+                    // close the fd rather than leaving the inherited one in place —
+                    // the workload gets EBADF, never the supervisor's stdin.
+                    let fd = unsafe { libc::open(b"/dev/null\0".as_ptr() as *const libc::c_char, libc::O_RDONLY) };
```
congwang-mk: **Bug: `Null` self-closes the target fd when the supervisor already closed that std fd (fail-closed guarantee violated).**

This arm does `let fd = open("/dev/null"); if fd<0 || dup2(fd,0)<0 { close(0) } if fd>=0 { close(fd) }`. If the supervisor was started with fd 0 closed (common for daemonized embedders), `open` returns the lowest free fd = `0` itself. `dup2(0,0)` is a POSIX no-op returning `0` (not `<0`), so the error branch is skipped; then `if fd>=0 { close(fd) }` runs `close(0)`, closing the very fd that was supposed to *be* `/dev/null`. The workload then gets `EBADF` on stdin instead of `/dev/null` — exactly the hardening path the comment promises.

The identical defect is in the stdout (line 1602, fd 1) and stderr (line 1622, fd 2) arms. The same `dup2(fd,fd)` no-op also hits the `Piped` arms: a pipe end opened `O_CLOEXEC` and allocated onto a pre-closed target fd keeps `CLOEXEC` set, so the stream vanishes at `execve`. Tests miss this because the cargo harness always has 0/1/2 open.

Fix: guard the `source == target` case — skip the `close`, and for the piped case clear `CLOEXEC` on the target.

crates/sandlock-core/src/sandbox.rs
```diff
@@ -1483,14 +1568,70 @@ impl Sandbox {
                 unsafe { libc::dup2(source_fd, target_fd) };
             }
 
-            if let Some((_, ref stdout_w)) = stdout_r {
-                unsafe { libc::dup2(stdout_w.as_raw_fd(), 1) };
+            // stdin (fd 0): Piped → child reads the pipe's read end; Null → /dev/null.
```
congwang-mk: **Cleanup: the three stdin/stdout/stderr arms duplicate the security-sensitive fail-closed sequence.**

The arms are copy-paste differing only in `O_RDONLY`/`O_WRONLY` and target fd `0/1/2`; the open/dup2/close fail-closed block is written three times (~55 lines of inline `unsafe`). The suite already needed `test_popen_stderr_piped` specifically to guard a copy-paste fd-1-vs-2 swap. A `fn wire_stdio(mode, target, pipe_end, open_flags)` (or a 3-row table) collapses this into one audited path — and the fix for the `Null`/`dup2(fd,fd)` bug above then lands in one place instead of three.

congwang-mk (commented):
Automated review of the streaming-stdio `popen` change. The refactor is clean and capture/inherit behavior is preserved byte-for-byte; three findings below (one correctness, one cleanup, one test flake).

crates/sandlock-core/tests/integration/test_popen.rs
```diff
@@ -0,0 +1,158 @@
...
+            .unwrap();
+        let stdout = File::from(child.take_stdout().expect("stdout pipe"));
+        let mut line = String::new();
+        BufReader::new(stdout).read_line(&mut line).unwrap();
+        gc_pid = line.trim().parse().expect("grandchild pid");
+        assert!(unsafe { libc::kill(gc_pid, 0) } == 0, "grandchild should be alive");
+        // sb (and child borrow) dropped here → Sandbox::drop kills the group.
+    }
+    // Give the kill + init-reap a moment to settle.
+    std::thread::sleep(std::time::Duration::from_millis(200));
```
congwang-mk: **Test flake: fixed 200ms sleep races zombie reaping.**

After the `Sandbox` drop SIGKILLs the group, the backgrounded `sleep 100` grandchild becomes a zombie until init/subreaper reaps it, and `kill(pid, 0)` returns `0` for an un-reaped zombie. On a loaded CI box where reaping takes >200ms, `assert!(!alive)` fails spuriously (and, less likely, a reused PID inside the window could mask a real leak). Prefer polling for `ESRCH` with a timeout over a single fixed sleep.

crates/sandlock-core/src/sandbox.rs
dzerik: Fixed at the root rather than the symptom (fde8d83). Instead of special-casing dup2(fd,fd), each piped source is now relocated to a fresh high fd via F_DUPFD_CLOEXEC (>=3) before being dup2'd down onto its 0/1/2 target, and the original CLOEXEC ends are mem::forget'd (they close at execve) so Drop can't close a fd we reassigned. That removes the src==target case — and the sibling stale-snapshot / cross-stream clobber hazards — as a class. dup2 is now checked and fails closed like Null. Coverage: the relocate path runs in every piped test e2e, plus a keystone unit test on relocate_high (returns >=3, CLOEXEC, same description). The exact closed-std-fd trigger isn't deterministically reproducible (the control PipePair claims the low fds first), so I left that as a documented limitation rather than a flaky test.

crates/sandlock-core/src/sandbox.rs
dzerik: Done — collapsed into one wire_child_stdio helper plus relocate_high, so the fail-closed / fd logic lives in a single audited path. (fde8d83)

crates/sandlock-core/tests/integration/test_popen.rs
dzerik: Fixed — replaced the fixed 200ms sleep with a poll for ESRCH (up to 5s), since the grandchild is a zombie until init reaps it and kill(pid,0) returns 0 for a zombie. (fde8d83)

PR #124, opened 2026-07-03
feat(go): streaming-stdio Popen — Process with per-stream StdioMode (RFC #67)

dzerik:
Go ergonomic layer over the C ABI `sandlock_popen` (merged in #120), mirroring the FFI and Python bindings. Part of the RFC #67 popen series (core #118, FFI #120).

`Sandbox.Popen(stdio Stdio, cmd...) (*Process, error)` forks the confined child with per-stream stdio — the streaming counterpart of `Spawn`. Each stream set to `StdioPiped` is handed back on the `*Process` as an `*os.File` (`Stdin`/`Stdout`/`Stderr`); inherit/null streams leave it nil. The zero `Stdio` inherits all three (identical to `Spawn`).

- `StdioMode` (Inherit=0/Piped=1/Null=2) shares the stable ABI discriminants; an out-of-range mode is rejected before the child is spawned.
- `os.NewFile` takes ownership of each piped fd, so closing the file closes the fd — no leaks (via `Close`, the `SetFinalizer` reap, or the `*os.File`'s own finalizer). On a null handle the FFI leaves the out-fds at -1, so no stream is wrapped on the error path.
- `Wait` closes a still-open piped `Stdin` first to deliver EOF (so a reader child like `cat` exits), then waits and frees the handle. `Close` closes all piped streams regardless of handle state and reaps the child.
- No wait-timeout: unlike the Python binding, `Kill` (process-group signal by PID) already interrupts a blocked `Wait` from another goroutine, so the undrained-pipe hang has a built-in escape hatch.

Tests mirror the FFI/Python popen suites: stdout streaming, stdin/stdout roundtrip through `cat`, wait-closes-unclosed-stdin (watchdog), both stdout+stderr piped, null yields no stream, invalid mode rejected, Kill interrupts a blocked Wait, Close closes streams + reaps, finalizer reaps an abandoned process, Close-after-Wait. README documents `Popen`/`Stdio`. `-race` clean.

Refs #67

### DIFF
--- a/go/sandbox.go
+++ b/go/sandbox.go
@@ -285,6 +285,29 @@ type Result struct {
 	Stderr   []byte // captured standard error
 }
 
+// StdioMode selects how one of a Popen'd process's standard streams is wired.
+// The values are the stable ABI discriminants shared with the C/Rust core.
+type StdioMode uint32
+
+const (
+	// StdioInherit shares the parent's fd (the child writes to the same
+	// terminal/file). It is the zero value, so an unset stream inherits.
+	StdioInherit StdioMode = 0
+	// StdioPiped connects the stream to a pipe; Popen hands the caller the
+	// owning end as an *os.File on the returned Process.
+	StdioPiped StdioMode = 1
+	// StdioNull connects the stream to /dev/null.
+	StdioNull StdioMode = 2
+)
+
+// Stdio selects the wiring of a Popen'd process's three standard streams. The
+// zero value wires all three as StdioInherit, matching Spawn.
+type Stdio struct {
+	Stdin  StdioMode
+	Stdout StdioMode
+	Stderr StdioMode
+}
+
 // ChangeKind classifies a filesystem change observed during a dry run.
 type ChangeKind byte
 
--- a/go/sandlock_linux.go
+++ b/go/sandlock_linux.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"runtime"
 	"strings"
 	"sync"
@@ -799,6 +800,13 @@ type Process struct {
 	h       *C.sandlock_handle_t
 	pid     int
 	waiting bool // a Wait owns the handle; other handle ops must defer to it
+
+	// Caller-owned stdio for a process started by Popen. Each is non-nil only
+	// for a stream wired StdioPiped; it owns the pipe fd (closing the file
+	// closes the fd). Nil for Spawn'd processes and for inherit/null streams.
+	Stdin  *os.File
+	Stdout *os.File
+	Stderr *os.File
 }
 
 // Spawn forks the sandboxed child, installs the policy, and releases it to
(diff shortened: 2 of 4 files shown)


go/sandlock_linux.go
```diff
@@ -869,16 +945,26 @@ func (p *Process) Pid() int {
 // Pause/Resume), which signal the process group by PID, can run concurrently
 // and interrupt it. The waiting flag reserves exclusive use of the handle for
 // the duration, so no other handle operation aliases it.
+//
+// Wait closes a still-open piped Stdin to deliver EOF, so do not write to Stdin
+// from another goroutine while Wait runs — the write would race a closing fd.
```
congwang-mk, line 956: **Doc gap: nothing says the `Result` of a Popen'd process carries no captured output.** The Python binding documents this in both `popen` and `Process.wait`; here neither `Wait`'s doc nor the README's Popen bullet mentions it, and the README still describes `Wait` as returning a "captured Result". A user coming from `Run` will read `res.Stdout` and get an empty slice with no hint why (the bytes went to `p.Stdout`, possibly already drained). One sentence here and one in the README fixes it.

go/sandlock_linux.go
```diff
@@ -838,10 +846,78 @@ func (s *Sandbox) Spawn(cmd ...string) (*Process, error) {
...
+// Process without waiting — the streaming counterpart of Spawn. For each stream
+// set to StdioPiped, the matching Process field (Stdin/Stdout/Stderr) is an
+// *os.File the caller reads/writes while the child runs; inherit/null streams
+// leave it nil. The zero Stdio inherits all three (identical to Spawn).
+//
+// Deadlock warning (as with os/exec pipes): a piped Stdin is yours — close it
+// before Wait, or a child that reads to EOF (e.g. cat) never exits and Wait
+// blocks. Likewise drain a piped Stdout/Stderr before Wait, or a child that
+// fills the pipe buffer blocks on write. As an escape hatch, Kill from another
+// goroutine interrupts a blocked Wait; Close reaps the child and closes the
```
congwang-mk, line 869: **Kill-after-exit idempotency now differs across the three bindings.** FFI `sandlock_handle_kill` returns 0 for an already-exited process, Python `Process.kill()` is a silent no-op after `wait()`, but Go's `Kill()` returns `ErrNotRunning` once the handle is freed (`signal` checks `p.h == nil` before the ESRCH mapping can apply). Since this doc actively promotes "Kill from another goroutine" as the escape hatch, the watchdog-races-normal-exit case (`Wait` returns just as the watchdog fires `Kill`) yields an error in Go where the other two bindings say success. Pre-existing `signal` semantics, but worth aligning while the contract is being written down: `Kill` on an already-reaped `Process` should arguably be nil too.

go/sandbox.go
```diff
@@ -285,6 +285,29 @@ type Result struct {
...
+	// StdioPiped connects the stream to a pipe; Popen hands the caller the
+	// owning end as an *os.File on the returned Process.
+	StdioPiped StdioMode = 1
+	// StdioNull connects the stream to /dev/null.
+	StdioNull StdioMode = 2
+)
+
+// Stdio selects the wiring of a Popen'd process's three standard streams. The
+// zero value wires all three as StdioInherit, matching Spawn.
+type Stdio struct {
```
congwang-mk, line 305: **Test gap: the zero-`Stdio`-equals-`Spawn` contract is stated in three places (here, the `Popen` doc, the README) but never exercised.** `TestPopenNullStdoutNoStream` is the closest test. A cheap `Popen(sandlock.Stdio{}, ...)` test asserting all three stream fields are nil and the child runs to success would pin the ABI default (`StdioInherit == 0`) that the zero-value contract relies on.

dzerik:
Thanks — addressed all three:

- **Kill idempotency**: `Kill()` on an already-reaped Process now returns nil, like ESRCH. But only the reaped case (`h == nil`) collapses to nil — I kept `pid <= 0` on a live handle as an error, so a genuine no-op kill isn't reported as success (and `killpg(-0)` stays unreachable). So the "Kill from another goroutine" escape hatch is race-free: a Kill firing just as Wait reaps the child returns nil, while a real failure still surfaces. Added a Kill-after-Wait test.
- **Doc**: `Wait`'s doc and the README now state the Result carries the exit status only — piped output went to `Stdout`/`Stderr` and inherited/null output to the parent fd or `/dev/null`, so (unlike `Run`) `Result.Stdout`/`Result.Stderr` are always empty. Qualified it so it's accurate for the inherit/null cases too, not just piped.
- **Zero-Stdio test**: added `TestPopenZeroStdioInheritsAllLikeSpawn` — `Popen(Stdio{}, ...)` asserts all three stream fields are nil and the child runs to success, pinning the `StdioInherit == 0` zero-value contract.

Force-pushed; CI green.

Issue #121, opened 2026-07-02
--net-allow breaks GUI apps


Dreadeath:
Hi, since 0.8.4 with --net-allow option I cannot start any GUI app. Running in Arch Linux, KDE.

congwang-mk:
Hi @Dreadeath 

This is expected, because technically Unix sockets don't belong to networking, so it is now gated by filesystem instead. So you need to add the Unix socket path used by your KDE as writable, for example `-w "$XDG_RUNTIME_DIR"`

Dreadeath:
Should it work with '-w /'? It doesn't.

And in first comment I meant that if there is no any --net-allow options, I can start GUI app.

congwang-mk:
Sorry for misunderstanding.

I just tried Unix socket with and with -w or --net-allow, I didn't see any unexpected issue. So, I guess your GUI app might be using something else like TCP loopback? Is there any way you could narrow it down?

dzerik:
The `-w /` not helping is actually the key clue — it rules out pathname unix sockets (those a filesystem grant *would* cover) and points at two things `-w` can't reach on KDE:

1. **TCP loopback** — as you suspected. Parts of Qt/KDE (and X11 when `DISPLAY` is a TCP display like `localhost:0`) connect to `127.0.0.1`. With `--net-allow` set you're in default-deny allowlist mode, so loopback is blocked unless it's listed.
2. **Abstract unix sockets** — D-Bus (which KDE leans on heavily) and often X11 (`@/tmp/.X11-unix/X0`) use *abstract-namespace* sockets: the address starts with a NUL byte (shown as `@`), so it has **no filesystem path at all**. `-w /` grants filesystem paths, so it can't authorize an abstract socket — which is exactly why `-w /` made no difference.

A quick way to pin which one it is:

```
strace -f -e trace=connect <gui-app>
```

and look at the `connect()` that returns `EACCES`/`EPERM`:
- `sun_path="@/..."` (or a leading `\0`) → **abstract** unix socket
- `sin_addr=127.0.0.1` → **TCP loopback**
- a real path like `/run/user/1000/bus` → **pathname** unix socket

Fix per case: loopback → add `127.0.0.1` to `--net-allow`; pathname socket → `-w <that path>`; abstract socket → there's currently no grant for it (it isn't a path and isn't TCP), so today the only options are dropping `--net-allow` or, if applicable, turning the abstract-socket gate off.

Dreadeath:
[...]

PR #125, opened 2026-07-03
fix(net): pass through connected AF_UNIX sendmsg under a destination policy

dzerik:
## Problem

With any `--net-allow`/`--net-deny` rule active, a `sendmsg()` on an **already-connected `AF_UNIX` stream socket** fails with `ECONNREFUSED`. `connect()` and `recvmsg()` on the same socket succeed — only the outbound `sendmsg()` is rejected. Since libwayland and D-Bus write their protocol via `sendmsg` (for the fd-passing capability), the client connects, sends its request, and then hangs waiting for a reply the peer never received. This is the GUI breakage reported in #121 (native Wayland clients such as `rofi`).

`send()`/`sendto` were unaffected — they Continue connected sockets unconditionally — which is why the breakage looked selective.

Minimal repro (any live pathname unix socket):

```
sandlock run --net-allow 127.0.0.1 -w / -- python3 -c \
 'import socket; s=socket.socket(socket.AF_UNIX,socket.SOCK_STREAM); s.connect("/run/user/1000/wayland-0"); print("send",s.send(b"x")); print("sendmsg",s.sendmsg([b"x"]))'
# send 1
# ConnectionRefusedError: [Errno 111]   <-- sendmsg only
```

## Root cause

In `sendmsg_on_behalf`, the connected / non-IP-family "Continue" fast path is gated behind `if !dest_policy`. With a destination policy active, a connected `AF_UNIX` send skips that path, is dup'd, and reaches `query_socket_protocol`, which returns `None` for a unix socket (no IP protocol) — treated as an unknown/forbidden protocol and refused with `ECONNREFUSED`. `sendmmsg_on_behalf` has the same gap.

## Fix

The IP destination policy governs IP sockets only; a `--net-allow` rule must not change how `AF_UNIX` sockets are handled. An `AF_UNIX` send is already fully decided by the named-unix fs gate (a named target is taken on-behalf; a connected/abstract send has no destination to validate). Probe `SO_DOMAIN` and Continue for unix sockets before the IP path, in both `sendmsg_on_behalf` and `sendmmsg_on_behalf`. This mirrors `sendto`, which only takes IP-family destinations on-behalf.

IP enforcement is untouched: non-unix sockets still resolve their protocol and validate every destination. Verified end-to-end that connected `AF_UNIX` `sendmsg` passes through under `--net-allow` (Wayland registry roundtrip restored) while denied IP destinations are still blocked.

## Test

Adds an integration test: a connected `AF_UNIX` stream `sendmsg` under an active `net_allow` must pass through and return the byte count.

### DIFF
--- a/crates/sandlock-core/src/network.rs
+++ b/crates/sandlock-core/src/network.rs
@@ -1082,25 +1082,29 @@ fn sendmsg_named_unix_on_behalf(
     sun_path: &std::path::Path,
     writable: &[std::path::PathBuf],
 ) -> NotifAction {
-    match send_named_unix_msghdr(notif, notif_fd, sockfd, msghdr_ptr, flags, sun_path, writable) {
-        Ok(n) => NotifAction::ReturnValue(n as i64),
+    match send_named_unix_msghdr(notif, notif_fd, sockfd, msghdr_ptr, sun_path, writable) {
+        Ok((dup_fd, m)) => {
+            let blocking = wants_blocking(dup_fd.as_raw_fd(), flags);
+            resolve_send(dup_fd, m, flags, blocking)
+        }
         Err(errno) => NotifAction::Errno(errno),
     }
 }
 
-/// Core of the named-unix on-behalf `sendmsg`: resolve+verify `sun_path`, copy
-/// the message's iovecs/control from the child, and send to the pinned inode
-/// via `/proc/self/fd`. Returns the bytes sent or an errno. Shared by the
-/// single-message `sendmsg` path and the per-entry `sendmmsg` path.
+/// Core of the named-unix on-behalf `sendmsg`: resolve+verify `sun_path` and
+/// copy the message's iovecs/control from the child, addressed to the pinned
+/// inode via `/proc/self/fd`. Returns the dup'd socket and a [`MaterializedMsg`]
+/// (which keeps the inode pin alive) for the caller to send — inline, and
+/// deferred if it would block. Shared by the single-message `sendmsg` path and
+/// the per-entry `sendmmsg` path.
 fn send_named_unix_msghdr(
     notif: &SeccompNotif,
     notif_fd: RawFd,
     sockfd: i32,
     msghdr_ptr: u64,
-    flags: i32,
     sun_path: &std::path::Path,
     writable: &[std::path::PathBuf],
-) -> Result<isize, i32> {
+) -> Result<(OwnedFd, MaterializedMsg), i32> {
     let pinned = match resolve_named_unix_target(notif.pid, sun_path, writable) {
         Ok(fd) => fd,
         Err(NotifAction::Errno(e)) => return Err(e),
@@ -1120,53 +1124,32 @@ fn send_named_unix_msghdr(
     let iovlen = (msg_iovlen as usize).min(1024);
     let iov_bytes = read_child_mem(notif_fd, notif.id, notif.pid, msg_iov_ptr, iovlen * 16)
         .map_err(|_| libc::EIO)?;
-    let mut data_bufs: Vec<Vec<u8>> = Vec::with_capacity(iovlen);
-    for i in 0..iovlen {
-        let off = i * 16;
-        if off + 16 > iov_bytes.len() {
-            break;
-        }
-        let base = u64::from_ne_bytes(iov_bytes[off..off + 8].try_into().unwrap());
-        let len = u64::from_ne_bytes(iov_bytes[off + 8..off + 16].try_into().unwrap()) as usize;
-        if len > MAX_SEND_BUF {
-            return Err(libc::EMSGSIZE);
-        }
-        if base == 0 || len == 0 {
-            data_bufs.push(Vec::new());
-            continue;
-        }
-        data_bufs.push(read_child_mem(notif_fd, notif.id, notif.pid, base, len).map_err(|_| libc::EIO)?);
-    }
-    let mut local_iovs: Vec<libc::iovec> = data_bufs
-        .iter()
-        .map(|b| libc::iovec {
-            iov_base: b.as_ptr() as *mut libc::c_void,
-            iov_len: b.len(),
-        })
-        .collect();
-
+    let data = flatten_iovecs(notif, notif_fd, &iov_bytes, iovlen)?;
     // Named target is always AF_UNIX, so translate SCM_RIGHTS / reject creds.
-    let (control_buf, _scm_fds) =
+    let (control_buf, scm_fds) =
         materialize_control(notif, notif_fd, msg_control_ptr, msg_controllen, true)?;
 
     let dup_fd = crate::seccomp::notif::dup_fd_from_pid(notif.pid, sockfd)
         .map_err(|e| e.raw_os_error().unwrap_or(libc::EBADF))?;
 
-    let mut msg: libc::msghdr = unsafe { std::mem::zeroed() };
-    msg.msg_name = &sun as *const libc::sockaddr_un as *mut libc::c_void;
-    msg.msg_namelen = sun_len;
-    msg.msg_iov = local_iovs.as_mut_ptr();
-    msg.msg_iovlen = local_iovs.len();
-    if let Some(ref ctrl) = control_buf {
-        msg.msg_control = ctrl.as_ptr() as *mut libc::c_void;
-        msg.msg_controllen = ctrl.len();
+    // The destination is the `/proc/self/fd/<pinned>` sockaddr; `pinned` must
+    // stay open (and at the same fd number) for that path to resolve, so the
+    // message keeps it alive. Copy the sockaddr bytes it currently encodes.
+    let addr = unsafe {
+        std::slice::from_raw_parts(&sun as *const libc::sockaddr_un as *const u8, sun_len as usize)
     }
-    let ret = unsafe { libc::sendmsg(dup_fd.as_raw_fd(), &msg, flags) };
-    if ret >= 0 {
-        Ok(ret)
-    } else {
-        Err(unsafe { *libc::__errno_location() })
-    }
+    .to_vec();
+
+    Ok((
+        dup_fd,
+        MaterializedMsg {
+            data,
+            control: control_buf,
+            addr,
+            _scm_fds: scm_fds,
+            _pinned: Some(pinned),
+        },
+    ))
 }
 
 /// Read a `sendmmsg` entry's `msg_name` and return its NAMED `AF_UNIX` path, or
@@ -1216,22 +1199,52 @@ fn sendmmsg_named_unix_on_behalf(
             // stop here and report a short send rather than passing it through.
             None => break,
         };
-        match send_named_unix_msghdr(notif, notif_fd, sockfd, entry_ptr, flags, &path, writable) {
-            Ok(n) => {
-                let bytes = (n as u32).to_ne_bytes();
-                let _ = write_child_mem(
-                    notif_fd,
-                    notif.id,
-                    notif.pid,
-                    entry_ptr + MSG_LEN_OFFSET as u64,
-                    &bytes,
-                );
-                sent += 1;
-            }
+        let (dup_fd, m) = match send_named_unix_msghdr(notif, notif_fd, sockfd, entry_ptr, &path, writable) {
+            Ok(pair) => pair,
             Err(errno) => {
                 first_errno = Some(errno);
                 break;
             }
+        };
+        // Non-blocking per entry so a batch never occupies the notification loop;
+        // the two cases that must leave the loop are deferred (see below).
+        let blocking = wants_blocking(dup_fd.as_raw_fd(), flags);
+        let ret = send_materialized_at(dup_fd.as_raw_fd(), &m, 0, flags | libc::MSG_DONTWAIT);
+        if ret >= 0 {
+            if blocking && (ret as usize) < m.data.len() {
+                // Partial stream on a blocking socket: complete this entry off the
+                // loop and report it, so a caller ignoring per-entry msg_len isn't
+                // silently truncated.
+                return complete_batch_entry(
+                    dup_fd, m, flags, ret as usize, notif_fd, notif.id, notif.pid,
+                    entry_ptr + MSG_LEN_OFFSET as u64, sent,
+                );
+            }
+            let bytes = (ret as u32).to_ne_bytes();
+            let _ = write_child_mem(
+                notif_fd, notif.id, notif.pid, entry_ptr + MSG_LEN_OFFSET as u64, &bytes,
+            );
+            sent += 1;
+        } else {
+            let err = unsafe { *libc::__errno_location() };
+            if err == libc::EAGAIN || err == libc::EWOULDBLOCK {
+                if sent == 0 && blocking {
+                    // Entry 0 would block entirely: a blocking socket never returns
+                    // EAGAIN, so complete it off the loop.
+                    return complete_batch_entry(
+                        dup_fd, m, flags, 0, notif_fd, notif.id, notif.pid,
+                        entry_ptr + MSG_LEN_OFFSET as u64, 0,
+                    );
+                }
+                // i>0 with nothing sent is a contract-legal short count; a
+                // non-blocking child at entry 0 gets EAGAIN.
+                if sent == 0 {
+                    first_errno = Some(libc::EAGAIN);
+                }
+                break;
+            }
+            first_errno = Some(err);
+            break;
         }
     }
     if sent > 0 {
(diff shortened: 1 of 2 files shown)


crates/sandlock-core/src/network.rs
```diff
@@ -1332,6 +1353,19 @@ async fn sendmsg_on_behalf(
...
+    // already fully decided by the named-unix fs gate above — a named target
+    // was taken on-behalf; a connected or abstract send has no IP destination
+    // to validate — so pass it through. Without this, a connected AF_UNIX
+    // sendmsg reaches `query_socket_protocol` below, which returns `None` for a
+    // unix socket and refuses the send with ECONNREFUSED. That breaks every
+    // connected-socket sendmsg user (Wayland, D-Bus, anything using sendmsg for
+    // its fd-passing capability) the moment any --net-allow/--net-deny rule is
+    // set, even though the send targets no network destination. Mirrors the
+    // sendto path, which only takes IP-family destinations on-behalf.
+    if socket_is_unix(dup_fd.as_raw_fd()) {
```
congwang-mk: **TOCTOU: this `Continue` under an active destination policy is racy and, I think, can't be made safe.** On `RESP_CONTINUE` the kernel re-resolves `arg0` (and `msg_name`) against the *live* child after this check. A malicious child can race `dup2(inet_sock, sockfd)` right after the `SO_DOMAIN` probe, so the kernel runs the send on an IP socket to a denied destination. That's exactly the "never Continue under dest_policy" invariant (lines 1314-1318 / 1609-1614) being broken; no `Continue`-based fix closes it, since both fd and `msg_name` are child-controlled at execution time.

The actual defect is narrow: `query_socket_protocol` is required eagerly and hard-fails `ECONNREFUSED` just below, but `protocol` is only consumed in the non-connected IP branch of `send_msghdr_on_behalf`. Suggested redirection, staying on the TOCTOU-safe on-behalf path:

1. Make `protocol: Option<Protocol>`, require it only in the `!connected` branch.
2. Translate `SCM_RIGHTS`: parse the cmsgs, `pidfd_getfd` each child fd into the supervisor, rewrite with supervisor fd numbers, close after send.

Step 2 is needed either way: on-behalf sends copy the control buffer raw, so fd-passing is already silently corrupted here and in `send_named_unix_msghdr`. Doing it on-behalf closes the race *and* fixes that gap in one place. Heads up that the repro and the new test only exercise the no-fd case, so neither approach is verified against an actual fd-passing `sendmsg`.

dzerik:
You're right on the TOCTOU — `Continue` under a destination policy can't be made safe, since the kernel re-resolves both `sockfd` and `msg_name` against the live child after our check, so a racing `dup2(inet_sock, sockfd)` redirects the send. Reworked to stay on the on-behalf path exactly as you suggested:

- `send_msghdr_on_behalf` now takes `protocol: Option<Protocol>` and consumes it only in the `!connected` IP branch. A connected send (every AF_UNIX send that reaches here — its connection was gated at connect time) needs no protocol and goes out on the immune `dup_fd`; a non-connected send whose protocol can't be resolved still fails closed with `ECONNREFUSED`. No `Continue` under a destination policy in either `sendmsg_on_behalf` or `sendmmsg_on_behalf`.
- `SCM_RIGHTS` translation: `translate_scm_rights` `pidfd_getfd`s each fd into the supervisor and rewrites its number before the send, on both the IP and named-unix on-behalf paths. As you noted, the control buffer was being copied verbatim, so fd-passing was silently corrupted — this closes that too.

Your heads-up that the repro/test only exercised the no-fd case was the useful nudge: I verified end-to-end that a real `SCM_RIGHTS` send now delivers a working fd (the receiver reads the passed file's exact bytes), and added that as an integration test.

Hardened the parser against a hostile child while I was in there:
- `cmsg_len` is bounds-checked against the *remaining* buffer (no `off + cmsg_len` overflow); a malformed header fails closed with `EINVAL` instead of panicking / scanning out of bounds. Unit-tested with a `usize::MAX`-ish `cmsg_len`.
- `SCM_CREDENTIALS` is rejected with `EPERM` rather than forwarded — on the on-behalf path the supervisor is the sender, so passing the child's crafted `pid/uid/gid` would let a privileged supervisor forge credentials the child can't assert. Failing closed is the safe choice.
- An oversized control buffer (> 16 KiB) fails closed with `EMSGSIZE` rather than being truncated into a partial cmsg chain.

IP enforcement is unchanged — non-unix sockets still resolve their protocol and validate every destination (verified: a denied UDP `sendmsg` is still blocked). Force-pushed; CI green.

crates/sandlock-core/src/network.rs
```diff
@@ -1497,11 +1600,23 @@ async fn send_msghdr_on_behalf(
...
+        // truncating (which could drop SCM_RIGHTS fds or send a malformed tail).
+        if msg_controllen as usize > MAX_CONTROL_BUF {
+            return Err(libc::EMSGSIZE);
+        }
+        match read_child_mem(notif_fd, notif.id, notif.pid, msg_control_ptr, msg_controllen as usize) {
+            Ok(raw) => {
+                let (buf, fds) = translate_scm_rights(notif.pid, &raw)?;
+                (Some(buf), fds)
+            }
+            Err(_) => (None, Vec::new()),
```
congwang-mk: Should this be return Err(EIO)? Otherwise it would silently send the message unexpectedly?

crates/sandlock-core/src/network.rs
```diff
@@ -1447,6 +1547,9 @@ async fn send_msghdr_on_behalf(
             None => return Err(libc::EAFNOSUPPORT),
         };
         let dest_port = parse_port_from_sockaddr(&addr_bytes);
+        // A non-connected IP send must have a resolved protocol to key the
+        // per-protocol allowlist. If it couldn't be resolved, fail closed.
+        let protocol = protocol.ok_or(ECONNREFUSED)?;
```
congwang-mk, line 1808: Should we check it against AF_UNIX so that translate_scm_rights() would be called only for Unix socket?

crates/sandlock-core/src/network.rs
```diff
@@ -1497,11 +1600,23 @@ async fn send_msghdr_on_behalf(
         });
     }
 
-    let control_buf = if msg_control_ptr != 0 && msg_controllen > 0 {
-        let len = (msg_controllen as usize).min(4096);
-        read_child_mem(notif_fd, notif.id, notif.pid, msg_control_ptr, len).ok()
+    // Copy the control buffer and rewrite any SCM_RIGHTS fds from child to
+    // supervisor numbers; `_scm_fds` keeps the fetched fds open across the send.
+    let (control_buf, _scm_fds) = if msg_control_ptr != 0 && msg_controllen > 0 {
```
congwang-mk: This code is duplicated with the above in send_named_unix_msghdr(), make it a helper?

congwang-mk (commented):
Thanks for your quick followup. Here are some additional issues:

congwang-mk:
NOT caused by this PR, but a related issue found by AI is:

```
A blocking on-behalf send can stall the entire supervisor loop

Seccomp notifications are processed sequentially on a single loop
(notif.rs: "Notifications are processed sequentially (not spawned)").
handle_net never returns NotifAction::Defer, so the on-behalf sends in
network.rs (libc::sendmsg in send_msghdr_on_behalf and
send_named_unix_msghdr, libc::sendto in sendto_on_behalf) execute
inline on that loop.

dup_fd_from_pid (pidfd_getfd) shares the open file description, so the
supervisor inherits the child's blocking mode. If the child's socket is
blocking and the send buffer is full (peer not draining), the
supervisor blocks indefinitely: no timeout applies (DEFER_TIMEOUT only
covers deferred work), and every trapped syscall from every sandboxed
process hangs until the peer reads or closes.

This predates PR #125 (reachable via connected TCP sendmsg under a
destination policy), but #125 amplifies it: connected AF_UNIX stream
sendmsg (Wayland, D-Bus) now routes through this path, so a stalled
peer freezes the whole sandbox, and a child can trigger it with no
network permission at all (fill one end of a socketpair via untrapped
write(), then sendmsg).

Nothing is bypassed; the impact is denial of service of enforcement.

Suggested fix: run the on-behalf send as NotifAction::Defer (bounds a
wedge to one defer slot and 30s). Faithful version: force MSG_DONTWAIT;
on EAGAIN with a blocking child, defer and await writability on the
dup'd fd (AsyncFd) before retrying, preserving blocking semantics
without occupying the loop.
```

Please let me know if you want to fix it together here. (If not, I can fix it separately.)

Thanks!

dzerik:
Thanks — addressed the inline comments and took the supervisor-loop issue on here (happy to do it together).

**Inline:**
- **Silent send on unreadable control** → now returns `EIO`. Folded the duplicated control handling of `send_msghdr_on_behalf` and `send_named_unix_msghdr` into one `materialize_control()`, which fails closed (`EIO` on an unreadable buffer, `EMSGSIZE` on an oversized one) instead of sending without the cmsgs.
- **Translate only for unix sockets** → `materialize_control` now translates `SCM_RIGHTS` / rejects `SCM_CREDENTIALS` only when the socket is `AF_UNIX` (probed via `SO_DOMAIN`); an IP socket's control (e.g. `IP_PKTINFO`) passes through untouched.
- **Dedup** → done, via that shared helper.

**Supervisor loop (blocking on-behalf send):** you're right, and it's a real DoS — a blocking child sending to a peer that stops draining wedged the whole sequential loop, reachable with no net permission at all, and #125 widened it to connected AF_UNIX (Wayland/D-Bus). Went with the faithful version:

- Each send is materialized into an owned message (flattened payload + translated control + inode pin + sockaddr), so it can be replayed off the loop.
- The first attempt is `MSG_DONTWAIT` on the loop, so it never blocks there. For a blocking child whose message didn't fully fit, it returns `NotifAction::Defer` — the deferred future awaits writability via `AsyncFd` (epoll, never blocking a worker thread) and pushes the rest, advancing a byte offset past each partial send until the full length is delivered, so a blocking `send(N)` still returns `N`. Control/address are attached only at offset 0, so `SCM_RIGHTS` transmits once. A non-blocking child gets the single-attempt result (short count / `EAGAIN`), as the kernel would give it.
- `sendmmsg` goes non-blocking per entry and stops the batch on a would-block or partial-stream entry (kernel non-blocking-sendmmsg semantics), rather than deferring mid-batch.

Verified end-to-end: a single blocking `send()` of 2 MiB to a peer that stalls 2 s then drains returns the full 2 MiB in ~2 s (not truncated to one `SO_SNDBUF`), and a concurrent thread's sends stay at ~1 ms latency while that send is stalled — the loop is free. Integration test added for the full-delivery case.

One heads-up on tradeoffs I landed on: a deferred send is bounded by the supervisor's deferred-work timeout, so a child blocked on a genuinely-stuck peer past that bound gets `EIO` rather than blocking forever — bounding the DoS. And `SCM_CREDENTIALS` on the on-behalf path is rejected with `EPERM` (from the earlier round) rather than forwarded, since the supervisor is the sender and can't faithfully assert the child's credentials. Happy to adjust either if you'd prefer different semantics.

crates/sandlock-core/src/network.rs
```diff
@@ -1617,32 +1860,46 @@ async fn sendmmsg_on_behalf(
...
                 Err(errno) => {
                     first_errno = Some(errno);
                     break;
                 }
+            };
+            // A batch sends each entry non-blocking so it never occupies the
+            // notification loop. A would-block entry stops the batch: the child
+            // sees the short count (or EAGAIN if the first entry blocked) and
+            // retries, matching the kernel's non-blocking sendmmsg semantics.
+            let ret = send_materialized_at(dup_fd.as_raw_fd(), &m, 0, flags | libc::MSG_DONTWAIT);
```
congwang-mk, line 1970: This batch (and the named-unix one) breaks blocking-socket semantics in two ways: if entry 0 would block, a blocking child gets EAGAIN with sent==0, which the kernel never returns on a blocking socket; and a partially-sent stream entry is counted as sent with a short msg_len, i.e. silent truncation, since blocking callers do not check per-entry msg_len.

A would-block at entry i>0 with nothing sent needs no fix: returning the short count i is contract-legal and callers retry. Only two cases must leave the loop, and they are the same operation ("complete entry i off-loop, then report i+1"):

1. entry 0 would block entirely
2. entry i partially sent (stream)

Suggested shape: split defer_send into a byte-level core, e.g. `async fn push_until_done(afd, m, offset) -> Result<usize, i32>` (resolve_send wraps it as today). Then, for a blocking child hitting case 1 or 2, return `NotifAction::defer` of: push_until_done, write the full msg_len back, return i+1. On error: Errno if i==0 and offset==0, else ReturnValue(i). Everything captured (notif copy, notif_fd, the MaterializedMsg) is already static-friendly. All three batch loops can share that tail.

Please do not materialize the whole batch up front to make it replayable: vlen x MAX_SEND_BUF is the supervisor OOM the 64MB cap exists to prevent.

dzerik:
Done — implemented exactly the shape you described.

- Extracted the byte-level core `async fn push_until_done(dup_fd, m, flags, offset) -> Result<usize, i32>`; `resolve_send`/`defer_send` wrap it for the single-message path.
- Added a shared `complete_batch_entry` tail used by all three batch loops. Only the two cases you called out leave the loop: **(1)** entry 0 would block entirely (offset 0), **(2)** entry i partially sent a stream (offset > 0). It runs `push_until_done`, writes the entry's **full** `msg_len` back (so a caller ignoring per-entry `msg_len` isn't silently truncated), and reports `i+1`; on error `Errno` when `i==0 && offset==0`, else `ReturnValue(i)`.
- A would-block at entry i>0 with nothing sent stays a contract-legal short count `i` (no defer). Batches are still materialized one entry at a time — never the whole `vlen`.
- Added an integration test: a blocking `sendmmsg` of one 4 MiB message on a connected stream socket to a stall-then-drain peer returns `ret=1` with the full `msg_len` and delivers every byte.

While re-reviewing this I also fixed a related gap: the blocking-vs-defer decision now honors a per-call `MSG_DONTWAIT` — `wants_blocking(fd, flags) = O_NONBLOCK clear && !(flags & MSG_DONTWAIT)`. A child that keeps a blocking socket but passes `MSG_DONTWAIT` on one `send` now gets the immediate short count / `EAGAIN` instead of a deferred block-to-completion (verified both ways e2e).

One thing I left as your prescription said, flagging it in case you'd prefer otherwise: an entry-0 partial-then-hard-error (some bytes queued, then e.g. `ECONNRESET`) returns `ReturnValue(0)` per "else `ReturnValue(i)`" with `i==0`. That's a value the kernel's own `sendmmsg` never returns for `vlen>0`; happy to switch it to the errno (or write the partial `msg_len` and return 1) if you'd rather.

CI green.

crates/sandlock-core/src/network.rs
```diff
@@ -1231,27 +1366,19 @@ async fn sendto_on_behalf(
...
-
-        // 6. Return result
-        if ret >= 0 {
-            NotifAction::ReturnValue(ret as i64)
-        } else {
-            let errno = unsafe { *libc::__errno_location() };
-            NotifAction::Errno(errno)
-        }
+        let blocking = wants_blocking(dup_fd.as_raw_fd(), flags);
+        resolve_send(dup_fd, m, flags, blocking)
```
congwang-mk, line 1385: The defer rework misses one path: `sendto_named_unix_on_behalf()` still calls `libc::sendto()` inline on the notification loop, with the child's blocking mode inherited through the dup. A child with only fs-write grants (no net rule) can bind a datagram socket under a writable path, never read it, and send until the queue fills; the next blocking `sendto` wedges the whole loop, the same DoS this PR fixes elsewhere. It already has the pinned inode and copied data, so wrapping them in a `MaterializedMsg` (with `_pinned` set) and going through `resolve_send` like this branch should be all it takes.

crates/sandlock-core/src/network.rs
```diff
@@ -1391,27 +1530,237 @@ fn prescan_msghdr(
...
+    NotifAction::defer(async move {
+        match push_until_done(dup_fd, m, flags, offset).await {
+            Ok(n) if n == m_len => {
+                let bytes = (m_len as u32).to_ne_bytes();
+                let _ = write_child_mem(notif_fd, notif_id, notif_pid, msglen_addr, &bytes);
+                NotifAction::ReturnValue((prior_count + 1) as i64)
+            }
+            // Partial stream then hard error: entry not fully delivered, so it is
+            // not counted; the child retries it (and the rest).
+            Ok(_) => NotifAction::ReturnValue(prior_count as i64),
```
congwang-mk: On your open question: let's align with the kernel. A blocking stream `sendmsg` that errors after partial progress returns the partial count (the error surfaces on the next call), so `sendmmsg` writes the partial `msg_len`, counts the entry, and returns `i+1`; it never returns 0 for `vlen > 0`. The current shape can spin a caller forever (`sent += ret` makes no progress on `ret == 0`) and invites duplicate bytes if the child retries an entry whose bytes are already queued. So: merge the two `Ok` arms, write `n` back as `msg_len`, and return `prior_count + 1` for any `Ok(n)`.

congwang-mk (commented):
Thanks for your patience through all these rounds; the reworks landed well. Two remaining points inline.

Issue #66, opened 2026-05-26
RFC: Credential injection via the proxy


congwang-mk:
## Goal

Let a sandboxed process authenticate to upstream APIs without the credential ever entering the sandbox's address space, environment, or filesystem. The real secret lives only in the supervisor; the child sees at most a per-session phantom token that authenticates nothing.

Related: #65 (transactional pipelines). Both lean on the same supervisor + proxy infrastructure.

## Why this is worth building

The standard pattern today is `OPENAI_API_KEY=sk-... agent`. Once the key is in the agent's env, it leaks via:

- Debug prints (`print(os.environ)`, loggers that capture process context).
- Tracebacks that include local-variable values.
- Child processes inheriting env (`subprocess.run(["curl", ...])`, `npm install` running install scripts).
- Prompt injection convincing the LLM to "print your environment."
- Exfiltration to an allowed-but-unintended endpoint (paste services, gists).
- Core dumps, memory snapshots, suspended VMs.

Every one of these is a real incident class. None are reachable if the secret is never in the agent process. That is the structural property worth chasing: the secret cannot leak from a place it does not exist.

Concrete scenarios this unlocks:

- AI coding agents with API access where the agent can run arbitrary shell commands.
- CI/CD running untrusted external PRs against internal APIs.
- IDE plugins / extension hosts calling APIs on the user's behalf.
- Multi-tenant agent serving where one user's key must never reach another user's workload.
- Running unfamiliar code from a README that wants `HF_TOKEN` or similar.
- MCP servers exposed to prompt injection.

## End state (what done looks like)

A user runs:

```bash
sandlock run --workdir . --service openai -- agent
```

That single flag arranges:

1. The real OpenAI key is read from the configured credential source (keystore / env / file / 1Password) into the supervisor's heap, wrapped in a zeroize-on-drop type.
2. A per-session phantom token (`sandlock_phantom_openai_<random>`) is generated.
3. `OPENAI_API_KEY=<phantom>` is set in the child env so SDKs that demand a non-empty env var work without modification.
4. The child's HTTPS calls to `api.openai.com` are intercepted by the existing seccomp `connect()` redirect, terminated at the local proxy via MITM with `--http-ca`, the real Bearer is injected, and the request is re-encrypted upstream.
5. Every credential use is recorded in the structured audit log by name, never by value.

[...]

PR #127, opened 2026-07-06
feat(net): credential injection primitives — inject an auth secret in the proxy (RFC #66)

dzerik:
Implements RFC #66, Phase 1, Layer 2 (transparent mode). This is the bottom-up primitive slice — the generic mechanism, not the `--service openai` presets (those are a later slice).

## What it does

A named secret lives **only in the supervisor**. When the sandboxed child makes a request to a matching upstream, the secret is rendered into an auth header (or query parameter) inside the MITM proxy — **after** `http_acl_check`, so a denied request never touches the secret — and the child never sees the real value.

```
sandlock run \
  --http-allow "* api.openai.com/*" --http-inject-ca \
  --credential openai=env:OPENAI_API_KEY \
  --http-inject "* api.openai.com/* bearer openai" \
  -- agent ...
```

## Design

- **Secret containment.** `SecretString` is zeroed on drop (volatile writes), has a redacted `Debug`, and deliberately implements no `Display`/`ToString`/`serde::Serialize` and is unreachable through any `Error` chain — so a stray `format!("{}", ..)` or a serialized policy can't re-open the leak the type exists to close. It flows to the proxy as `Arc<Vec<InjectRule>>` and is `#[serde(skip)]` on the sandbox (re-loaded from source per build, never persisted).
- **ACL before inject.** `AclService::handle` applies the first matching rule only after the ACL passes; the deny path returns before any rendering. Audit records the credential *name*, never the value.
- **`OnExistingHeader` defaults to `Replace`.** The proxy owns the credential, and every major SDK requires an API key to be set and always sends its own placeholder `Authorization`. Keeping it would forward the placeholder and never inject — so the rule's target header is replaced by default; pass a trailing `add-only` to keep a child-set value instead.
- **Sources.** `env:` / `file:` / `fd:` (a trailing newline is stripped); `literal:` is rejected (leaks via ps / shell history). `file:`/`fd:` are capped at 64 KiB; `fd:` refuses the std streams and reads through a `dup` so the caller's fd stays open. Env-sourced secrets are stripped from the child's environment post-fork, so the agent can't read the value straight out of its own env.

## Safety rails

- `--http-inject` without an ACL proxy is rejected; without a CA it warns that only plaintext HTTP is intercepted (a rule for an HTTPS host would silently no-op otherwise).
- A secret that can't render into a header (e.g. contains CRLF) fails the request rather than forwarding it uncredentialed.
- A checkpoint is rejected while injection is configured, since the non-serialized secrets can't be restored into the image.

## Auth shapes

`bearer` · `basic:<user>` (base64) · `header:<name>` / `apikey:<name>` · `query:<param>` (raw bytes percent-encoded, so a binary key isn't corrupted).

## Tests

- **Unit:** `SecretString` redaction, source loading (incl. `literal:`/std-fd rejection and the 64 KiB cap), every `AuthShape`, the `Replace` default vs `add-only`, query `AddOnly` de-dup, a CRLF secret failing the request, and `resolve_inject_rules` Arc de-dup + env-strip collection.
- **Integration:** a credential is injected into the upstream while the child carries none; a **denied** request never renders the secret (403, upstream never contacted); an `env:` credential is scrubbed from the child.
- HTTPS/MITM injection runs the same scheme-agnostic `apply` path; a TLS end-to-end test needs a CA-trust harness and is tracked as a follow-up (noted in the test file).

## Deferred to later slices

`--service openai` (Layer 1) presets, phantom-token env swap, explicit proxy mode, body-level injection, and a structured audit log (the current audit is a by-name stderr line).

### DIFF
--- a/crates/sandlock-core/src/context.rs
+++ b/crates/sandlock-core/src/context.rs
@@ -501,6 +501,12 @@ pub(crate) fn confine_child(args: ChildSpawnArgs<'_>) -> ! {
     for (key, value) in &sandbox.env {
         std::env::set_var(key, value);
     }
+    // Remove env vars whose value was loaded into the supervisor as a credential
+    // source, so the agent can't read the real secret straight from its own
+    // environment (this is the child; the mutation is process-local and post-fork).
+    for name in &sandbox.inject_env_strip {
+        std::env::remove_var(name);
+    }
 
     // 13b. GPU device visibility
     if let Some(ref devices) = sandbox.gpu_devices {
--- a/crates/sandlock-core/src/context.rs
+++ b/crates/sandlock-core/src/context.rs
@@ -498,14 +498,17 @@ pub(crate) fn confine_child(args: ChildSpawnArgs<'_>) -> ! {
             std::env::remove_var(&key);
         }
     }
-    for (key, value) in &sandbox.env {
-        std::env::set_var(key, value);
-    }
     // Remove env vars whose value was loaded into the supervisor as a credential
     // source, so the agent can't read the real secret straight from its own
     // environment (this is the child; the mutation is process-local and post-fork).
+    // This runs *before* applying `sandbox.env`, so the inherited real secret is
+    // dropped but a deliberate placeholder the user passes (e.g.
+    // `--env OPENAI_API_KEY=dummy`, which SDKs need set to start) still survives.
     for name in &sandbox.inject_env_strip {
         std::env::remove_var(name);
+    }
+    for (key, value) in &sandbox.env {
+        std::env::set_var(key, value);
     }
 
     // 13b. GPU device visibility
--- a/crates/sandlock-core/src/credential.rs
+++ b/crates/sandlock-core/src/credential.rs
@@ -205,6 +205,18 @@ impl std::fmt::Debug for InjectRule {
     }
 }
 
+/// What [`InjectRule::apply`] did on the `Ok` path — so the caller logs a
+/// truthful audit line instead of claiming "injected" when an `add-only` rule
+/// actually left the caller's own credential in place.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Applied {
+    /// The secret was rendered into the request.
+    Injected,
+    /// `add-only` and the request already carried the target header/param, so
+    /// the caller's value was kept and no secret was written.
+    Skipped,
+}
+
 impl InjectRule {
     /// True if this rule applies to the given request.
     pub fn matches(&self, method: &str, host: &str, path: &str) -> bool {
@@ -214,16 +226,17 @@ impl InjectRule {
     /// Render the secret into `parts` (header or query), honoring `on_existing`.
     /// The rendered header is marked sensitive so downstream logging redacts it.
     ///
-    /// `Err(())` means the secret could not be rendered (e.g. it contains bytes
-    /// illegal in an HTTP header value) — the caller must reject the request
-    /// rather than forward it with no (or a partial) credential. A no-op skip
-    /// under `AddOnly` (the agent already set the header) is `Ok`.
+    /// `Ok(Injected)` wrote the secret; `Ok(Skipped)` left a caller-supplied
+    /// value in place under `add-only`. `Err(())` means the secret could not be
+    /// rendered (e.g. it contains bytes illegal in an HTTP header value) — the
+    /// caller must reject the request rather than forward it with no (or a
+    /// partial) credential.
     ///
     /// Zeroing is best-effort: [`SecretString`] itself is wiped on drop, but the
     /// transient buffers built here (the `Bearer `/`Basic ` byte vecs, the base64
     /// string, the query pair) hold a copy of the secret and are dropped without
     /// volatile zeroing, so a copy briefly lives in freed heap.
-    pub fn apply(&self, parts: &mut hyper::http::request::Parts) -> Result<(), ()> {
+    pub fn apply(&self, parts: &mut hyper::http::request::Parts) -> Result<Applied, ()> {
         match &self.auth {
             AuthShape::Bearer => {
                 let mut v = b"Bearer ".to_vec();
@@ -248,18 +261,18 @@ impl InjectRule {
         parts: &mut hyper::http::request::Parts,
         name: &str,
         value: &[u8],
-    ) -> Result<(), ()> {
+    ) -> Result<Applied, ()> {
         let hn = hyper::header::HeaderName::from_bytes(name.as_bytes()).map_err(|_| ())?;
         if self.on_existing == OnExistingHeader::AddOnly && parts.headers.contains_key(&hn) {
-            return Ok(()); // agent already set it — leave it, not a failure
+            return Ok(Applied::Skipped); // agent already set it — leave it, not a failure
         }
         let mut hv = hyper::header::HeaderValue::from_bytes(value).map_err(|_| ())?;
         hv.set_sensitive(true);
         parts.headers.insert(hn, hv);
-        Ok(())
+        Ok(Applied::Injected)
     }
 
-    fn set_query(&self, parts: &mut hyper::http::request::Parts, param: &str) -> Result<(), ()> {
+    fn set_query(&self, parts: &mut hyper::http::request::Parts, param: &str) -> Result<Applied, ()> {
         let uri = &parts.uri;
         let path = uri.path();
         let existing = uri.query();
@@ -268,7 +281,7 @@ impl InjectRule {
         if self.on_existing == OnExistingHeader::AddOnly {
             if let Some(q) = existing {
                 if q.split('&').any(|kv| kv.starts_with(&prefix)) {
-                    return Ok(());
+                    return Ok(Applied::Skipped);
                 }
             }
         }
@@ -298,7 +311,7 @@ impl InjectRule {
             b = b.authority(a.clone());
         }
         parts.uri = b.path_and_query(new_pq).build().map_err(|_| ())?;
-        Ok(())
+        Ok(Applied::Injected)
     }
 }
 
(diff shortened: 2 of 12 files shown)


crates/sandlock-core/src/context.rs
```diff
@@ -501,6 +501,12 @@ pub(crate) fn confine_child(args: ChildSpawnArgs<'_>) -> ! {
     for (key, value) in &sandbox.env {
         std::env::set_var(key, value);
     }
+    // Remove env vars whose value was loaded into the supervisor as a credential
+    // source, so the agent can't read the real secret straight from its own
+    // environment (this is the child; the mutation is process-local and post-fork).
+    for name in &sandbox.inject_env_strip {
```
congwang-mk, line 507: This strip runs after the `sandbox.env` set loop, so it also deletes an explicitly passed placeholder (`--env OPENAI_API_KEY=dummy`). SDKs refuse to start without the var set, which is exactly the placeholder flow the Replace default is designed for, so the canonical `env:` workflow breaks before any request is made. Strip the inherited env first, then apply `sandbox.env`: the inherited real secret is still removed and deliberate user config is honored.

crates/sandlock-core/src/credential.rs
```diff
@@ -0,0 +1,620 @@
...
+        hv.set_sensitive(true);
+        parts.headers.insert(hn, hv);
+        Ok(())
+    }
+
+    fn set_query(&self, parts: &mut hyper::http::request::Parts, param: &str) -> Result<(), ()> {
+        let uri = &parts.uri;
+        let path = uri.path();
+        let existing = uri.query();
+        // Honor AddOnly: don't append a param the request already carries.
```
congwang-mk, line 288: `Replace` does not replace for the query shape: the existing param is only checked under `AddOnly`, so the default path appends a duplicate (`key=child&key=sk-real`, as the test pins). Most frameworks read the first occurrence, so the upstream authenticates against the child's placeholder and returns 401; it is also parameter-pollution ambiguity. On `Replace`, filter out existing `param=` pairs before appending, mirroring `headers.insert`.

crates/sandlock-core/src/credential.rs
```diff
@@ -0,0 +1,620 @@
...
+        Some((k, a)) => (k, Some(a)),
+        None => (spec, None),
+    };
+    let shape = match (kind, arg) {
+        ("bearer", None) => AuthShape::Bearer,
+        ("basic", Some(user)) if !user.is_empty() => AuthShape::Basic { username: user.to_string() },
+        // `apikey:<header>` and `header:<name>` are the same rendering.
+        ("header" | "apikey", Some(name)) if !name.is_empty() => AuthShape::Header { name: name.to_string() },
+        ("query", Some(param)) if !param.is_empty() => AuthShape::Query { param: param.to_string() },
+        _ => {
```
congwang-mk, line 347: A username containing `:` is accepted here (`basic:a:b` yields username `a:b`), which shifts the credential boundary in the encoded `user:pass`: the upstream sees user `a`, password `b:<secret>`. RFC 7617 forbids `:` in the user-id; reject it at parse time.

crates/sandlock-core/src/credential.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+}
+
+/// Parse an `AuthShape` from the auth token of an `--http-inject` rule:
+/// `bearer` | `basic:<user>` | `header:<name>` | `apikey:<name>` | `query:<param>`.
+/// Returns `(shape, credential_name)`.
+pub fn parse_auth(spec: &str, credential: &str) -> Result<AuthShape, SandboxError> {
+    let (kind, arg) = match spec.split_once(':') {
+        Some((k, a)) => (k, Some(a)),
+        None => (spec, None),
```
congwang-mk, line 331: Stale doc: says it returns `(shape, credential_name)` but the function returns only the `AuthShape`.

congwang-mk (commented):
Design looks solid overall: ACL-before-inject is structural, SecretString containment is well closed, and reusing HttpRule keeps one matching semantics. Four inline findings below. Two smaller notes not worth inline threads: (1) with a CA configured, a rule matching a port-80 host still injects into cleartext with no warning (the build-time warning only fires when no CA is set); (2) refusing fd:0 blocks the common stdin secret pattern (docker/systemd pipes), consider allowing it or documenting the <(...) workaround.

crates/sandlock-core/src/transparent_proxy/service.rs
```diff
@@ -150,6 +154,33 @@ impl AclService {
 
         let (mut parts, body) = req.into_parts();
         parts.uri = uri;
+
+        // ACL passed: attach a credential if a rule matches. First match wins.
```
congwang-mk, line 214: Do we need to consider multi-header injection? Or is it out of scope?

congwang-mk:
Another potential issue: without CA, we get HTTP port 80, this might not be expected by this injection?

congwang-mk:
A minor name issue: `--http-inject` is a bit similar to the existing `--http-inject-ca`, maybe rename it to `--http-auth` (or any other better name)?

dzerik:
Thanks for both notes. Status plus one design question for you.

**`--http-inject` → `--http-auth`**: agreed — will rename to disambiguate from `--http-inject-ca`.

**Cleartext (port 80) injection** — this is the real call, and I'd rather not over-restrict it, because plaintext HTTP *at the app layer* is legitimately secure in a common deployment: a service-mesh sidecar (Istio/Linkerd mTLS) terminates TLS, so the app speaks plain `http` to `127.0.0.1`/the sidecar while the wire is mTLS. Refusing cleartext injection outright would break that.

`service.rs` already knows the scheme per request (`"https"` for the MITM path, `"http"` for plaintext), so we can act on it at request time rather than only warning at build time. Options, your call on surface:

- **(a)** warn-at-request-time, still inject — no new flag; warn once per host when injecting over `http`. Closes the "silent in the with-CA case" gap and keeps the mesh use case working.
- **(b)** refuse cleartext by default + opt-in `--http-auth-allow-cleartext` — fail-closed, but adds a flag.
- **(c)** refuse cleartext unconditionally — no flag, but breaks the mesh-mTLS case above.

I lean **(a)** (minimal surface, keeps legit cleartext working, makes the leak visible instead of silent), but it's your surface — happy to wire (b) if you'd rather fail-closed.

**Multi-header (`service.rs:158`)**: intentionally first-match-wins, one credential per request. Multiple simultaneous auth headers seemed unusual, so I kept it out of scope — glad to add it if you have a case in mind.

congwang-mk:
Emitting  a warning on CLI side is fine, how about API side? :)

congwang-mk:
BTW, while you are on it, please update README.md and docs/sandbox-reference.md together in this PR.

dzerik:
Pushed the three follow-ups:

1. **Renamed `--http-inject` → `--http-auth`** (and the `SandboxBuilder::http_auth` field/method to match), so it no longer reads as a sibling of `--http-inject-ca`. The CA flags (`--http-inject-ca` / `--http-ca`) are untouched.

2. **The cleartext warning is now API-side.** You're right that a CLI-only warning misses library callers — so it lives in the transparent proxy (`transparent_proxy/service.rs`), which is core and runs for every caller. It fires at request time (the only place the scheme is known), latched once per run, when a credential is actually injected over plaintext `http`. It names only the credential and host — never the secret value — and is a warning only, the request still proceeds.

   I kept the existing build-time warning as well: it covers a different risk (a CA isn't set, so credentials scoped to an HTTPS host would be silently dropped), whereas the new one fires when a credential actually goes out in cleartext.

3. **Docs** — a README example plus a `Credential injection` section in `docs/sandbox-reference.md`, in this PR.

crates/sandlock-core/src/credential.rs
```diff
@@ -0,0 +1,654 @@
...
+    // once and several rules can share the secret), and collect the env-var
+    // names of `env:` sources so the child can be denied them — otherwise the
+    // agent would just read the value straight from its own environment.
+    let mut loaded: HashMap<&str, Arc<SecretString>> = HashMap::new();
+    let mut env_strip: Vec<String> = Vec::new();
+    for p in &parsed {
+        if !loaded.contains_key(p.name) {
+            loaded.insert(p.name, Arc::new(load_secret(p.source)?));
+            if let Some(var) = p.source.strip_prefix("env:") {
+                if !env_strip.iter().any(|v| v == var) {
```
congwang-mk: If a credential is defined but not referenced, this would leak it to processes in the sandbox?

dzerik:
Pushed 00b83b8. While addressing your env-strip question I ran a deeper self-audit of the whole feature and found a more serious issue I want to flag directly.

**Critical — cross-origin credential exfiltration (now fixed).** The proxy keyed the ACL check, `verify_host`, and the credential match on the URI-authority host (`host`), but rebuilt the outbound request from the Host header (`host_hdr`) — and the upstream client routes by that. So a hostile child could send `GET https://api.openai.com/v1` with `Host: attacker.example`: the secret is injected for the allowed host, then the request (carrying `Authorization: Bearer <secret>`) is forwarded to `attacker.example`. That's cross-origin exfiltration of the exact credential the feature promises the child never sees, plus an egress-ACL bypass, in both plaintext and MITM modes. Worth flagging that the root — `host` vs `host_hdr` divergence and forwarding by the Host header — predates this PR in the transparent proxy; credential injection just turns it into a secret leak, so it may deserve a look independent of this feature. Fixed here by failing closed when the URI-authority host and the Host header disagree (`split_host_rejected`), before verify/ACL/inject; origin-form requests are unaffected. Unit test added.

**Your env-strip question — confirmed, fixed.** You're right: `env_strip` was built only from *referenced* credentials, so a declared-but-unreferenced `--credential X=env:SECRET` stayed in the child's environment. It's now stripped for every declared `env:` credential regardless of reference. Two things fall out of the same root:
- it also left the secret in the child's memory → the on-disk **checkpoint** image (`capture.rs` guards on `inject`, which is empty in the unreferenced case) — the strip-all fix closes that too;
- the **supervisor** still holds the secret in its own `/proc/<pid>/environ`, reachable by a child granted `/proc` with no PID-ns. That one is a deeper property of the supervisor having to hold the secret at all, so I'm flagging it rather than fixing it here.

**Two smaller ones while in there:** `file:` credentials had no protection analogue (env: stripped, fd: dup-closed) — now a build-time warning if the secret file sits inside an fs-read grant. And `apply()` returned `Ok(())` for both a real injection and an add-only no-op, so the audit line claimed "injected" when it actually kept the child's header — it now returns Injected/Skipped and logs truthfully.

Left multi-header out of scope as before (first-match-wins). Happy to adjust any of these.

dzerik:
Correction on my previous comment — the split-host fix I described was incomplete; a self-audit found two live bypasses. Pushed a proper fix (c4c3fd6).

The earlier guard compared `uri.host()` with a `split(':')` view of the Host header, but the outbound request re-parses the full Host header as an authority (whose `.host()` strips userinfo), so security-host and forward-host could still diverge — origin-form, and `Host: allowed:x@attacker.example` (userinfo), both still exfiltrated. Replaced with a single `request_authority()` that parses both the Host header and any URI authority as `http::uri::Authority`, rejects userinfo and a URI-vs-Host mismatch, and drives verify/ACL/inject **and** the outbound request from that one value — so the host a credential is matched for is exactly the host it's sent to, in every request form. Fixes IPv6 literals for free.

The same audit closed three more while I was in there: the `file:` overlap warning now covers `fs_writable` (write access includes read), `fs_mount` host paths, and the chroot root, not just `fs_readable`; query Replace/AddOnly now match the param name, so a value-less `?key` is the target rather than appending `key&key=secret`; and `fd:0` (stdin) is allowed per your earlier note (only `fd:1`/`fd:2` refused), with `--http-auth` rejecting >5 tokens.

Also confirmed the supervisor-side `/proc/<pid>/environ` residual I mentioned isn't reachable: `handle_proc_open`'s per-PID filter already returns EACCES for any pid not in the sandbox ProcessIndex, and the supervisor is never registered.

Issue #65, opened 2026-05-26
RFC: Multi-sandbox transactions for pipelines


congwang-mk:
## Goal

Promote `commit` / `abort` from a per-sandbox property to a first-class property of a `Pipeline` (or any group of sandboxes that should succeed or fail together). Stages should be able to see each other's in-flight writes under a defined isolation model, and the runtime should commit all of them together at the end or none of them at all.

This is an RFC for the full feature, not a single PR. It describes the end state, the design choices, the recommended defaults, and a delivery phasing.

## What exists today

- `Sandbox` owns `cow_branch: Option<Box<dyn CowBranch>>` with `commit() / abort() / cleanup() / changes()` (`crates/sandlock-core/src/cow/mod.rs`).
- Only 1 backend now:
  - **seccomp**: pure userspace, no mount, no privilege. Supervisor intercepts opens, plans copy-on-write into an upper directory, tracks deletions in memory. Auto-enabled when `fs_isolation = None` and a workdir is set. Works on any Linux 5.9+.
- Per-sandbox `on_exit` / `on_error` policy (`BranchAction::{Commit, Abort, Keep}`) auto-applied in `Sandbox::Drop` (`crates/sandlock-core/src/sandbox.rs:1615`).
- `Pipeline` forks N stages with inter-stage pipes (`crates/sandlock-core/src/pipeline.rs`), zero awareness of COW. Each stage's `Drop` independently decides commit/abort based on its own exit code.
- Dry-run for a single sandbox (`Sandbox::do_dry_run`) captures `changes()` then aborts.
- `Gather` (fan-in) exists for data flow over pipes; it has no COW story.

## The end state (what \"done\" looks like)

A user can declare a group of sandboxes (initially a `Pipeline` or `Gather`, eventually any user-defined set) and run them under a transaction. The runtime guarantees:

1. Every stage observes prior stages' writes according to the chosen isolation model (default: read-committed for sequential, snapshot for parallel siblings).
2. At end of transaction, either every stage's writes commit to the underlying workdir, or none of them do.
3. Parallel siblings that touch the same path produce a structured conflict, surfaced to the caller with a configurable resolution policy.
4. The pipeline can be dry-run: the union of intended changes is reported and nothing is committed.
5. The transaction works on any sandlock-supported kernel (5.9+) without elevated privilege.
6. Crash mid-commit is recoverable: a subsequent sandlock invocation can either complete or roll back the interrupted transaction from durable transaction metadata.
[...]

PR #144, opened 2026-07-15
fix(cow): return ENOENT when opening a file deleted in the branch

dzerik:
## Problem

A file deleted in a seccomp COW branch is recorded as a whiteout in the in-memory `deleted` set; the real lower file is intentionally left untouched until commit. The stat/access handlers honor that whiteout and return `ENOENT`, but the **read/open path did not**: for a deleted path opened without `O_CREAT`, `prepare_open` returned `CowOpenPlan::Skip`, which `handle_cow_open` maps to `NotifAction::Continue` — so the kernel opened the untouched lower file and the reader saw the file's **pre-delete content**.

The two paths disagreed, so a deletion was invisible to a reader.

### Example

```sh
# workdir/secret.txt exists with content "PREDELETE", COW active
rm -f secret.txt
test -e secret.txt        # -> false   (stat path: ENOENT, correct)
dd if=secret.txt of=out   # -> succeeds, out == "PREDELETE"  (open path: leaks lower bytes)
```

`cat secret.txt` happens to hide the bug because coreutils `cat` stats first and short-circuits on the (correct) stat `ENOENT`; a bare `open()` with no preceding stat (`dd`, a raw `openat`, some readers) hits the leaking path.

## Fix

Add `CowOpenPlan::Deleted` for a whiteout opened without `O_CREAT`; `handle_cow_open` maps it to `NotifAction::Errno(libc::ENOENT)`, matching the stat/access handlers. Reads of a deleted file now fail closed instead of leaking lower content.

## Tests

- Unit: `prepare_open` reports `Deleted` (not `Skip`) for a whiteout opened `O_RDONLY`.
- Integration: a child deletes a workdir file, then a bare open (`dd`) must be **denied** (`ENOENT`) and copy **zero** bytes — the test fails with `leaked: "PREDELETE"` on the old behavior.

## Scope

This is a general COW whiteout fix on the read/open path — it also reproduces in a single (non-pipeline) sandbox. It additionally makes a stage's deletion visible to later stages under the read-committed model of the multi-sandbox transactions work (#65), which is why I split it out as its own small change.

### DIFF
--- a/crates/sandlock-core/src/cow/seccomp.rs
+++ b/crates/sandlock-core/src/cow/seccomp.rs
@@ -386,7 +386,12 @@ impl SeccompCowBranch {
             if flags & O_CREAT != 0 {
                 return self.ensure_cow_copy(&rel).map(Some);
             }
-            return Ok(None);
+            // Whiteout: the lower file still physically exists with its
+            // pre-delete bytes. Surface the deletion so the caller returns
+            // ENOENT rather than falling through to the lower file — matching
+            // the async prepare_open (CowOpenPlan::Deleted) and the stat/access
+            // handlers.
+            return Err(BranchError::Deleted);
         }
 
         // O_EXCL: fail if file already exists (in upper or lower)
@@ -1405,6 +1410,22 @@ mod tests {
         // O_RDONLY
         let plan = branch.prepare_open(&path, 0).unwrap();
         assert!(matches!(plan, CowOpenPlan::Deleted));
+    }
+
+    #[test]
+    fn test_handle_open_read_deleted_reports_deleted() {
+        // Sync mirror of test_prepare_open_read_deleted_reports_deleted: the
+        // chroot dispatcher calls the sync handle_open, so a read-only open of a
+        // whiteout must surface BranchError::Deleted (mapped to ENOENT at the
+        // chroot call site) instead of Ok(None), which fell through to the
+        // untouched lower file and leaked its pre-delete bytes.
+        let (workdir, storage) = setup_workdir();
+        let mut branch = SeccompCowBranch::create(workdir.path(), Some(storage.path()), 0).unwrap();
+        branch.mark_deleted("existing.txt");
+        let path = abs(&branch, "existing.txt");
+        // O_RDONLY
+        let err = branch.handle_open(&path, 0).unwrap_err();
+        assert!(matches!(err, BranchError::Deleted));
     }
 
     #[test]
--- a/crates/sandlock-core/src/cow/seccomp.rs
+++ b/crates/sandlock-core/src/cow/seccomp.rs
@@ -46,6 +46,10 @@ pub enum CowCopyPlan {
 pub enum CowOpenPlan {
     /// No interception needed — let the kernel handle it.
     Skip,
+    /// The path was deleted in this branch (a whiteout) and is opened without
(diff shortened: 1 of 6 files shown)


crates/sandlock-core/src/cow/seccomp.rs
```diff
@@ -447,7 +451,11 @@ impl SeccompCowBranch {
             if flags & O_CREAT != 0 {
                 return self.prepare_cow_copy(&rel);
             }
-            return Ok(CowOpenPlan::Skip);
+            // Whiteout: the file was deleted in this branch. Do NOT skip to the
+            // lower file (which still physically exists with its pre-delete
+            // content); report the deletion so the caller returns ENOENT,
+            // matching the stat/access path.
+            return Ok(CowOpenPlan::Deleted);
```
congwang-mk, line 463: This fixes the seccomp dispatcher, but chroot mode still leaks: the sync `handle_open` above (~line 385) returns `Ok(None)` for a whiteout without `O_CREAT`, and `chroot/dispatch.rs` maps `Ok(None)` to a fall-through that opens the real lower file via `openat2_in_root`, so a reader still gets the pre-delete bytes there. `handle_open` should surface the whiteout too, with the chroot call site mapping it to ENOENT.

crates/sandlock-core/src/cow/seccomp.rs
```diff
@@ -447,7 +451,11 @@ impl SeccompCowBranch {
             if flags & O_CREAT != 0 {
                 return self.prepare_cow_copy(&rel);
```
congwang-mk, line 457: `O_CREAT` on a whiteout bypasses the fail-closed guarantee this PR adds: `prepare_copy` removes the path from `deleted` and copies the lower bytes into upper, so `rm f; touch f; cat f` (or a bare `open(f, O_CREAT|O_RDONLY)`) resurrects the pre-delete content. After unlink, `O_CREAT` should create an empty upper file instead of copying lower. Fine as a tracked follow-up if out of scope here.

congwang-mk (commented):
Thanks for the PR! Two issues in the same bug class, one per inline comment.

dzerik:
Fixed Inline #1: added `BranchError::Deleted` (the sync mirror of `CowOpenPlan::Deleted`) — `handle_open` now surfaces the whiteout instead of `Ok(None)`, and the single chroot call site maps it to `ENOENT`, matching the async path and the stat/statx/access/readlink handlers. Covers both `openat` and legacy `open` (legacy synthesizes into `handle_chroot_open`).

Added a chroot-mode integration test (`test_chroot_cow_read_deleted_file_is_enoent`) that drives a bare `cat` open (no preceding stat) of a whiteout through the real chroot dispatcher and asserts the pre-delete lower bytes do not leak — the chroot sibling of your async `test_seccomp_cow_read_deleted_file_is_enoent`. Confirmed it fails (leaks the lower `"PREDELETE"` bytes) without the `handle_open` change, so it locks the actual no-leak contract rather than just the branch return value.

Inline #2 (`O_CREAT`-on-whiteout resurrection) I'll take as the tracked follow-up you suggested — it also needs `test_quota_handle_open_create_denied` rewritten (it currently asserts the old copy-into-upper quota behavior), so keeping it out of this ENOENT-signal change keeps both reviewable in isolation.

PR #148, opened 2026-07-17
pipeline: transactional sequential pipelines over a shared COW workdir (RFC #65 Phase 1)

dzerik:
**Round 2 (revision `1ee26a3`):** all five round-2 concerns addressed; CI green 13/13; three decisions flagged in the comment below.

---

Transactional sequential pipelines over a shared COW workdir — the Rust-core slice of RFC #65.

Rewritten after your design review; the previous body described a different shape and is no longer accurate.

Fair warning on the size of this: your five items turned out to be more than I bargained for. Item 1 was one bug in one place, or so it looked — fixing it surfaced the same failure class inside the fix for item 3, and fixing *that* surfaced it again at a cancellation point one `await` deeper. Turtles all the way down, three levels of them, and each level looked closed until the next probe went in. That is why the diff is larger than a five-item review has any right to produce, and why a good part of it is tests.

## What it is now

A dedicated `Transaction` type (`crates/sandlock-core/src/transaction.rs`), not a second method on `Pipeline`. Stages run sequentially over one shared COW upper; the workdir is untouched until the commit, which is all-or-nothing.

```rust
Transaction::new([Stage::new(sb1, argv1), Stage::new(sb2, argv2)])
    .run(Some(Duration::from_secs(30))).await?;   // or .dry_run(..)
```

## Your five points

**1. `Drop` destroying data on a failed commit.** `commit()` now enters `Preserved(MergeInterrupted)` *before* the first destructive operation, so every `?` inside the merge returns with `Drop` disarmed and the upper intact. The blast radius of adding `Drop` to `SeccompCowBranch` is stated in its rustdoc rather than framed as a leak fix, and an explicit `BranchAction::Keep` is no longer overridden — an abandoned sandbox that asked to keep its branch keeps it (`set_keep_if_abandoned`).

Two more instances of the same failure class turned up while chasing this one, both fixed and both with tests that fail without the fix: the commit-lock timeout path returned `Err` while the branch was still `Open`, and — the one I would have missed by reading exit paths only — *cancelling* the future during the lock wait dropped an `Open` branch. Cancellation is not an exit path, so no code runs; the fix moves the whole commit phase onto `spawn_blocking` with the branch moved in, which makes the loss structurally impossible and also gets the walk + per-file copy + fsyncs off the async executor.

**2. Wrong type.** `run_transactional` is gone from `Pipeline`. `Transaction` has no `BitOr` and is built only from an explicit stage list, and `Pipeline::stages` is now private (`into_stages()`) so `Transaction::new(chain.stages)` no longer compiles — the footgun is closed by construction rather than by a doc comment. Its rustdoc states the execution model in the first paragraph: sequential, no inter-stage pipes, stdio inherited, data exchanged through the shared workdir.

I did not build `Transaction::new(sandboxes)` from the RFC's end state: a `Sandbox` carries no argv, so the coordinator needs `Stage { sandbox, args }`. If you want that name reserved for the eventual signature, say so and this becomes `Transaction::from_stages`.

**3. Not first-committer-win.** Correct, and the name is gone. What the lock provides is mutual exclusion between merges, stated as such in the code. It is now wait-with-timeout rather than fail-fast, and losing the wait is no longer `Ok(TxnOutcome { committed: false })` — it is `TxnError::Conflict`, so a caller that ignores the distinction cannot silently lose a pipeline's work. The stale-base lost update you described is unchanged and documented: this is concurrency safety, not serializable isolation.

**4. Phase 1 acceptance.** You are right that this does not close Phase 1, and I am not claiming it does — see the scope section below.

**5. Error typing.** `AbortReason` is an enum (`StageFailed { index, status }`, `TimedOut`, `DryRun`); `TxnError` separates the commit channel (`Branch`, `CommitLock`, `Conflict`, `Merge`, `CommitAbandoned`) with distinct exit codes, and flattens to `SandboxRuntimeError::Branch` rather than `Child(String)`. No string matching required to tell a stage failure from a commit failure from a conflict.

## Scope: this is Phase 1a

Still owed, not done here: `commit()`/`abort()` as a public surface (only `run`/`dry_run` ship), the Python and Go bindings, the Python end-to-end test the RFC's acceptance names, and tooling to *act* on a preserved branch — `list_preserved` discovers them, but there is no `recover()`/`discard()` primitive yet, so a preserved upper is currently the operator's problem.

`TxnOutcome` does carry the RFC's `changes` field, and `dry_run()` exists.

## Preserved storage

A preserved branch now records its deletions in the on-disk marker alongside the upper, because a recovery that replays only the upper would resurrect every file the transaction deleted — and `TxnError::Merge` tells the operator that recovering that storage *is* how the transaction gets finished. Two limits are documented rather than hidden: the deletion set is written at preserve time, so a recovery may re-apply a deletion that has since landed (a no-op), and there is no recovery command yet.

## Tests

`seccomp.rs` 64 → 106 unit tests, `transaction.rs` 27, `test_transaction.rs` 38, plus additions to `test_cow.rs`; suites on this branch are 599 lib / 342 integration.

Every test was required to fail against broken code before being counted — the ones that could not were rewritten or dropped. Two that shipped earlier in this branch turned out to pass vacuously (a lock-contention test whose contention had already expired, and a latency test whose only assertion was a wall-clock bound that a 20ms regression could not trip); both are fixed.

One caveat you will see in CI: `test_cow::test_seccomp_cow_exec_packed_argv_relocation` (be654d7, on `main`) flakes under compile load — roughly 2 runs in 3 here, clean in isolation. Not from this branch, but it will red this PR intermittently.

## Out of scope, filed separately

Writing the COW merge tests turned up four defects in the COW layer that predate this work: #158 (copy-up of a FIFO wedges the supervisor indefinitely), #159 (whiteouts are not recursive, so a deleted directory's contents stay readable and can be republished by `commit()`), #160 (renaming a lower-only directory silently destroys its contents), #161 (`rmdir` on a non-empty directory succeeds where the kernel gives `ENOTEMPTY`). None are touched here.

Refs #65.


dzerik:
Implemented the first-committer-win lock the body flagged as unimplemented (latest commit). Before the merge, `run_pipeline_txn` takes an exclusive non-blocking `flock` on the shared workdir; a transaction that finds it held loses the race as a **benign structured outcome** — `TxnOutcome { committed: false, abort_reason: Some("...lost the first-committer-win race...") }` with its upper aborted — rather than a hard error or a raced merge.

Scoped to the transactional commit path only; the single-sandbox `commit()` is unchanged (so `on_exit=Commit` doesn't silently drop a loser's changes). It serializes overlapping merges but does not detect a stale base, so two non-overlapping transactions touching the same file still resolve last-writer-wins — concurrency safety, not serializable isolation. Integration test holds the lock and asserts a fully-successful pipeline returns `committed=false` with nothing merged; verified it commits (races) without the lock.

congwang-mk:
Design review (no code details). Core shape is sound: coordinator owns one branch, stages attach, finalize on every exit path. Five things.

**1. The `Drop` backstop destroys data on a failed commit (blocking).**

`commit()` sets `finished = true` only at the end. Every `?` inside the merge loop returns `Err` with `finished` still false, so the branch drops and `cleanup()` removes the upper. On `ENOSPC` or `EACCES` partway through the merge, the workdir is left half-merged *and* the upper holding the unmerged remainder is deleted. Before this PR that upper survived and was recoverable.

This contradicts the doc comment on `run_transactional`: partial state is only tolerable because you can recover from the upper. It also forecloses RFC Phase 3, whose premise is a recovery sweep over durable branch state. Set `finished` (or a separate `merging` flag) at the *start* of `commit()`.

Separately, adding `Drop` to `SeccompCowBranch` changes behavior for every branch holder, not only transactions. That blast radius should be stated, not framed as a leak fix.

**2. `run_transactional` is on the wrong type.**

`Pipeline` is defined by `|`: N stages concurrent, stdout to stdin. `run_transactional` runs the same object sequentially with no pipes and inherited stdio. Same builder, two incompatible execution models selected by method name. `a | grep foo | c` works under `.run()` and hands `grep` the parent's stdin under `.run_transactional()`, with no error.

The RFC names `Transaction::new(sandboxes).run()` as the eventual surface and calls `Pipeline` "the first user of the underlying primitive." Since we take hard breaks pre-1.0, build that type now rather than let `Pipeline` carry two meanings through Phase 2.

**3. The commit lock is not first-committer-win.**

The `flock` covers only the merge. Txn A snapshots at open, B commits, A merges over B's result from a stale base. That is a lost update. The mechanism is "don't interleave two merges," which is what the follow-up comment says correctly but the code comment does not; "first-committer-win" reads as a guarantee it does not provide.

Losing the race also returns `Ok(TxnOutcome { committed: false, .. })`, so a caller that does not inspect `committed` silently loses a whole pipeline's work behind a success return. Given `LOCK_NB` and short merges, wait-with-timeout is the better default.

**4. Phase 1 acceptance is not met.**

RFC Phase 1 specifies `commit() / abort() / dry_run()` plus the Python surface, with acceptance requiring at least one Python end-to-end test. This ships only `run_transactional()`. `TxnOutcome` also drops the RFC's `changes` field, so there is no way to see what a transaction did without diffing the workdir, and there is no `dry_run` at all.

Slicing is fine, but then call it Phase 1a and say what is still owed. As written it reads as closing Phase 1. The body is stale regardless: it still says the workdir lock is unimplemented.

**5. Error typing undercuts the RFC's CLI goal.**

Every validation failure is `SandlockError::Runtime(SandboxRuntimeError::Child(...))`, so a misconfigured pipeline is indistinguishable from a child process error. `abort_reason` is `Option<String>`, so callers must string-match to tell timeout from stage failure from lost race. The RFC calls for an `AbortReason` enum and exit codes that separate stage failure from commit failure from conflict. Cheap now, expensive once the Python SDK parses these.

dzerik:
Flagging this since editing the description doesn't notify: I answered your five points by rewriting the PR body rather than replying inline. The shape changed enough that the old description was actively misleading — `run_transactional` is gone, the transactional path moved onto its own type, and the lock's contract is different.

Where each one landed:

1. `commit()` enters `Preserved(MergeInterrupted)` *before* the first destructive operation, so every `?` in the merge returns with `Drop` disarmed and the upper intact. Chasing it surfaced two more instances of the same class: the commit-lock timeout path returned `Err` with the branch still `Open`, and *cancelling* the future during the lock wait dropped an `Open` branch — cancellation is not an exit path, so no cleanup code runs. The commit phase now runs on `spawn_blocking` with the branch moved in, which makes that loss structurally impossible. The blast radius of adding `Drop` to `SeccompCowBranch` is stated in its rustdoc instead of being framed as a leak fix.

2. `run_transactional` is gone from `Pipeline`. `Transaction` is a separate type with no `BitOr`, built only from an explicit stage list, and `Pipeline::stages` is now private (`into_stages()`), so `Transaction::new(chain.stages)` no longer compiles — closed by construction rather than by a doc comment.

3. The name is gone. What the lock provides is mutual exclusion between merges, stated as such in the code. It is wait-with-timeout rather than fail-fast, and losing the wait is `TxnError::Conflict`, not `Ok(TxnOutcome { committed: false })`, so a caller that ignores the distinction cannot silently lose a pipeline's work. The stale-base lost update you described is unchanged and documented as a limit.

4. Agreed that this does not close Phase 1. It is scoped as Phase 1a with an explicit "still owed" list: public `commit()`/`abort()`, the Python and Go bindings, the Python end-to-end test the acceptance names, and tooling to act on a preserved branch. `TxnOutcome.changes` and `dry_run()` do ship.

5. `AbortReason` is an enum (`StageFailed { index, status }`, `TimedOut`, `DryRun`); `TxnError` separates the commit channel (`Branch`, `CommitLock`, `Conflict`, `Merge`, `CommitAbandoned`) with distinct exit codes and flattens to `SandboxRuntimeError::Branch` rather than `Child(String)`. No string matching to tell a stage failure from a commit failure from a conflict.

One question for you in the body: the RFC's `Transaction::new(sandboxes)` can't be the signature as-is, since a `Sandbox` carries no argv and the coordinator needs `Stage { sandbox, args }`. If you want that name reserved for the eventual surface, say so and this becomes `Transaction::from_stages`.

congwang-mk:
Design-level pass on the new shape. The five points from last round are all genuinely addressed, mostly by construction. Five new concerns, ranked:

**1. The commit lock doesn't cover single-sandbox commits.** Your own comment admits it: a plain `Sandbox` committing into the same workdir takes no lock, so it can interleave with a transaction merge and tear both. Either move the flock into `SeccompCowBranch::commit()` so both callers get it, or file it as a named follow-up before merge.

**2. Preservation into per-pid `$TMPDIR` undermines the durability story.** The error says "recovering that storage is how the transaction gets finished", but the default storage is `$TMPDIR/sandlock-cow-<pid>`: tmp-cleaners delete it and `list_preserved` can't sweep across dead pids' bases. Preserved work should land somewhere stable (Phase 3 already names `$XDG_RUNTIME_DIR`); at minimum, document that recovery-sensitive callers must set `fs_storage`.

**3. Inherited-only stdio won't survive contact with the SDK.** `AbortReason::StageFailed` carries index and status, nothing else, and `stages[].stdout/stderr` are always `None`. "Stage 2 exited 1" with no stderr forces every SDK caller into redirection boilerplate. Decide whether stderr gets captured per-stage before the Python binding freezes `TxnOutcome`'s shape.

**4. `committed: bool` + `abort_reason: Option<...>` is a tri-state held together by an invariant**, and `DryRun` as an "abort reason" for a successful run is strained. A `disposition: Committed | DryRun | Aborted(AbortReason)` enum makes invalid states unrepresentable. Cheap now, expensive after the bindings mirror it.

**5. The recovery surface leaks the backend.** Docs and links point at `crate::cow::seccomp::list_preserved`, but preserved-work recovery is conceptually backend-generic. Re-export through `transaction::` (or a `recovery` module) and document that path instead.

On your question: keep `Transaction::new(stages)`. A `Sandbox` carries no argv, so the RFC's `new(sandboxes)` was never going to be the real signature; `Stage { sandbox, args }` is the correct unit. No reservation needed, no `from_stages`.

Before merge I'd want a fix or a filed follow-up for 1 and 2, and a decision on 3 and 4. CI is green across all 13 checks, including the known-flaky COW test.

dzerik:
Round-2's five concerns are addressed in the latest revision; CI is green across all 13 checks.

Notes on the two I reworked hardest:

- **(1) commit lock** — the flock lives in `SeccompCowBranch::commit()` now, so a plain `Sandbox` and a `Transaction` both serialize on it; `finish_branch` no longer opens the workdir, so there's a single lock layer and no self-deadlock. Preserve-reason invariant: `CommitDeferred` iff the workdir was untouched (lock contended, or the workdir vanished before the merge), `MergeInterrupted` iff the merge ran and may have touched it — and a contended retry no longer downgrades a `MergeInterrupted` marker.
- **(3) per-stage stderr** — the live tee is an **independent** `O_NONBLOCK` description opened on `/proc/self/fd/2` (not a dup of fd 2), written best-effort with the byte dropped on `EAGAIN`/`EPIPE`, so a slow or closed parent stderr can never stall the pipe drain or fill the child's stderr pipe. After a stage is reaped its process group is killed before the drain, so a backgrounded descendant can't hold the pipe open.

Three calls I made that you may want to weigh in on rather than have me decide silently:

1. **A plain `Sandbox`'s commit takes the workdir flock, and it runs in `Drop`.** Under contention it blocks up to `DROP_COMMIT_LOCK_WAIT` (5s) then defers-and-preserves (recoverable via `list_preserved`). That's a new blocking destructor — happy to make the wait configurable/shorter, or a `Duration::ZERO` try-once, if you'd rather it never sit on the calling thread.
2. **Default preserved-work base is `$XDG_RUNTIME_DIR/sandlock-cow`** — durable across a user's dead pids, but session-scoped (logind reaps on last-session-exit). Documented that a daemon which must outlive its session sets `fs_storage`. Say if you'd prefer a session-independent default.
3. **The deferral logs via `eprintln!("sandlock: …")`**, matching the crate's existing best-effort diagnostics; `sandlock-core` has no `tracing`/`log` dep, so I didn't pull one in for one line.

`Transaction::new(stages)` stays per your call.

congwang-mk:
Thanks for the thorough round-3 update. All five round-2 concerns are genuinely addressed, and your three calls (the 5s Drop wait, the XDG default, eprintln) all land where I would have landed them. Keep them as they are.

Two things before merge:

**1. The preserved marker is not durable at the one moment it exists for.** `write_preserved_marker` is a plain `fs::write`, no fsync of the file or its directory entry, yet it is written immediately before the merge's first destructive step precisely so a crash mid-merge leaves something to find. On power loss the workdir can be half-merged while the marker never reached disk, and a branch dir without a marker is invisible to `list_preserved`. c483683 fsync'd `deleted.log` and its directory entry for the same reason; the marker deserves the same. Related pinhole: `pid=412` truncated to `pid=41` still parses as a complete record with a wrong pid, and pid liveness is what separates a live merge from a dead one. Requiring the trailing newline after the pid line closes it cheaply.

**2. The env-mutating unit test can flake the suite.** `create_falls_back_to_tmp_when_xdg_base_is_foreign` sets `XDG_RUNTIME_DIR`/`TMPDIR` process-wide while sibling tests in the same binary (`list_preserved_default_base_spans_pids`, `drop_cleans_undisposed_branch_but_keep_preserves`) call `create(None)` and read the same env under cargo's parallel runner. A branch created mid-mutation lands under the guard's tempdir and is deleted with it. CI is green today but this is latent; serialize them behind a shared env-lock mutex.

dzerik:
Both items are addressed and the branch is on top of main with #162 in it. Three things I would rather you decide than have me decide silently.

**How it was integrated.** Merge, not rebase (`190b82a`). The `DeletionSet` change is a semantic conflict rather than a textual one, so a rebase would have meant re-resolving it once per commit, and merging leaves `1ee26a3` — the head both of your review threads anchor to — intact. `cow/deletions.rs` is touched by six comment lines: no `remove()`, no signature change.

**The resolution.** The un-delete cases fold into `covers()` + upper presence, so the whiteout set stays append-only and durable. The retryable-remainder contract moved to a separate, RAM-only `applied_deletions`, surfaced through `outstanding_deletions()`, which feeds both the retry and the marker body. One deliberate divergence: `changes()` no longer surfaces a deletion that is both outstanding and re-created in the upper; a test pins that. If you would rather it were visible, the additive answer is a public accessor.

**Item 1, the marker.** `write_preserved_marker` is now temp → `write_all` → `sync_all` → `rename` → `sync_dir`, with the strict call at the head of `commit_merge` before the first destructive step, and `cleanup()` retiring the marker durably before the storage it describes. The truncated-record pinhole is closed by requiring the trailing newline, with a test that truncates at every offset and fails at byte 155 without it.

Making it durable surfaced three problems in the merge path itself, all of which are fixed here: an "applied" test that treated *any* stat error as "the path is gone", so an unstatable whiteout was counted as merged and the successful tail then removed the only record of it; a marker refresh gated on "an unlink happened" rather than "the outstanding set shrank", which let the copy phase publish a path the marker still named; and a directory fsync that opened without `O_DIRECTORY`/`O_NOFOLLOW` and unconfined, so a parent swapped for a FIFO hung the commit indefinitely while holding the workdir lock. Each has a test that fails without its fix. The same fsync is also tolerant now for the errnos where it is structurally impossible (`EACCES` on a `0o300` directory cannot be fsync'd by any means) and still refreshes the marker, since an unsynced removal is non-destructive while a stale marker is not.

**Item 2, the env-mutating test.** Closed by removing the mutation rather than guarding it: the storage-base resolution is now injectable and the test passes both directories in, so nothing touches process env. `grep set_var` in that file is empty and the two sibling tests are untouched — the named class went from 33/40 failures at `--test-threads=8` to 0/40. The shared mutex you suggested is still available if you would rather have it; other `set_var` callers remain in the same binary, so only the collision you named is closed. While measuring that, one more load-dependent test turned up (`ensure_secure_base_revalidates_on_create_race`, 13/30 failures under load, and `--test-threads=1` does not help it); it is replaced by driving the re-validation arm directly.

**One I did not fold in.** The pre-existing directory fsync further down has the same FIFO hazard and comes from main, so it is left untouched here rather than smuggled into this PR — happy to send it as a one-liner, or to include it if you prefer a single hunk.

This slice has been unusually productive at surfacing things: nearly every fix in it uncovered the same failure class one level deeper, which is why it grew well past the two items.

Issue #143, opened 2026-07-15
Chroot named-AF_UNIX gate bypass: lexical check + Continue (symlink/TOCTOU) and non-UTF-8 None-collapse (fail-open)


dzerik:
# Chroot named-AF_UNIX gate is bypassable: lexical check + `Continue` (symlink/TOCTOU) and non-UTF-8 `None`-collapse (fail-open)

**Component:** `crates/sandlock-core/src/network/` (connect / sendto / sendmsg / sendmmsg named-AF_UNIX gate)
**Affected tree:** `main` @ `c63cf0e`
**Impact:** an untrusted, unprivileged, non-ptracing child under an active `has_unix_fs_gate` policy can connect/send to a named AF_UNIX socket that is **not** under any fs-write grant — a full bypass of the only enforcement layer for pathname AF_UNIX (Landlock has no access right for unix connect, per `connect.rs:119-121`). Two distinct defects share this gate.

---

### Where
- `network/connect.rs:135-136` (connect)
- `network/send.rs:112-113` (sendto)
- `network/unix.rs:183-184` (sendmsg, via `unix_sendmsg_gate`)
- `network/send.rs:383` (sendmmsg chroot loop)
- Predicate: `network/verdict.rs:74-77` — `path_under_any` calls `chroot::resolve::confine` (`chroot/resolve.rs:7-28`), a purely lexical `.`/`..` collapse that never touches the filesystem and never follows symlinks.

### What
In chroot mode a named AF_UNIX target is validated with `path_under_any(&path, &ctx.policy.chroot_writable)` on the supervisor's immune copy, and on a match returns `NotifAction::Continue`:

```rust
if ctx.policy.chroot_root.is_some() {
    if path_under_any(&path, &ctx.policy.chroot_writable) {
        NotifAction::Continue          // <-- kernel re-reads sun_path and resolves it
    } else {
        NotifAction::Errno(libc::EACCES)
    }
} else {
    connect_named_unix_on_behalf(/* pins the real inode, never Continue */)
}
```

`Continue` hands the syscall back to the kernel, which **re-reads sun_path from child memory and performs real path resolution, following symlinks**. The non-chroot branch is safe by contrast: it pins the real inode with `O_PATH` under `/proc/<pid>/root`, checks `real_path_under_any`, and acts on-behalf against `/proc/self/fd/<pin>` — never `Continue`.

[...]

PR #154, opened 2026-07-19
fix(net): close a sendto non-IP destination-policy TOCTOU (sendto half of #130)

dzerik:
Splits the high-priority **sendto** half out of #130, as you suggested on that PR ("The sendto tightening is a high-priority fix... Maybe you want to separate it out as a fix which could land quickly"). The sendmsg/sendmmsg symmetry stays in draft #130; I'll rebase it onto this once this lands.

## The bug

A non-IP `sendto` with no fs-path gate fell through to `_ => NotifAction::Continue`, so the kernel re-read `sockfd` and the destination address *after* the supervisor's decision. Under an active destination policy a child could race a `dup2(inet_sock, sockfd)` + address swap into that window and ride the `Continue` out to a denied IP. (Same `Continue`-shape as the chroot AF_UNIX gate report, #143.)

## The fix

Under `has_net_destination_policy` the arm no longer `Continue`s: it pins the fd via `dup_fd_from_pid`, classifies on the **stable** socket domain of the pinned fd (via the new pure `classify_send_path`), and either sends the unix datagram on-behalf on that pinned fd or fails closed with `EAFNOSUPPORT`. With **no** destination policy there is nothing to bypass, so `Continue` is preserved unchanged.

The non-`UnixOnBehalf` outcomes collapse to a single fail-closed errno arm rather than `unreachable!()`, so any future drift in the classifier fails closed instead of panicking — a panic on this seccomp-notif path would unwind the supervisor task and DoS the sandbox.

## Testing

The pure `classify_send_path` helper + its 4 unit tests land here (sendto is its first consumer). `send_path_non_ip_on_non_unix_socket_is_rejected` is the deterministic **fail-without-fix** witness: pre-fix logic yielded `Continue` for that input, the helper now yields `Reject`/`EAFNOSUPPORT`.

I deliberately did **not** add an integration test for the fail-closed errno: `sendto(AF_INET dgram fd, AF_UNIX sockaddr)` returns `EAFNOSUPPORT` from the kernel too, so such a test is green pre- and post-fix and would be decorative. The bypass this closes is a race that isn't deterministically reproducible from a guest, so the pure-helper unit test is the honest witness.

`classify_send_path` is `pub(crate)` (no SDK/FFI surface). It is the shared classifier the sendmsg/sendmmsg rework in #130 consumes fully; landing it here with sendto as first consumer lets #130 become a pure consumer on rebase.

## Behavior scope (please review)

This is fail-closed hardening that only triggers **under a destination policy**, but it does change some behavior — flagging explicitly:

- A non-IP `sendto` on a **non-unix, non-IP** datagram family (`AF_PACKET`/`AF_VSOCK`/…) now returns `EAFNOSUPPORT` instead of `Continue`. (Narrow: `AF_NETLINK` is virtualized to a unix socket and stays on the on-behalf path; `AF_PACKET` needs `CAP_NET_RAW`.) `connect_on_behalf` still `Continue`s these families, so there is a connect-vs-sendto asymmetry.
- **Abstract / fs-gate-off named** unix datagrams now send **on-behalf**, so the receiver observes the supervisor's `SO_PEERCRED`/`SCM_CREDENTIALS` rather than the child's. Enabling an IP allowlist thus changes unix-datagram credential semantics (D-Bus/journald/etc.). This is intrinsic to closing the TOCTOU (Continue is unsafe once the fd can be swapped).
- Until #130 lands, `sendto` and `sendmsg` differ on a non-IP unix datagram (sendto sends on-behalf; sendmsg still returns `EAFNOSUPPORT`). This preserves sendto's pre-existing behavior (it already sent, via `Continue`) while closing its TOCTOU; #130 aligns sendmsg.

Happy to add a family carve-out (fail closed only for `AF_INET`/`AF_INET6` sockets) if you'd prefer to keep exotic non-unix datagram families working.

### DIFF
--- a/crates/sandlock-core/src/network/materialize.rs
+++ b/crates/sandlock-core/src/network/materialize.rs
@@ -11,6 +11,8 @@ use std::os::unix::io::{AsRawFd, OwnedFd, RawFd};
 use crate::seccomp::notif::read_child_mem;
 use crate::sys::structs::{SeccompNotif, AF_INET, AF_INET6};
 
+use super::verdict::DestShape;
+
 /// Maximum buffer size for sendto/sendmsg on-behalf operations (64 MiB).
 /// Prevents a sandboxed process from triggering OOM in the supervisor.
 pub(crate) const MAX_SEND_BUF: usize = 64 << 20;
@@ -218,6 +220,30 @@ pub(crate) fn named_unix_socket_path(addr_bytes: &[u8]) -> Option<std::path::Pat
         return None;
     }
     std::str::from_utf8(raw).ok().map(std::path::PathBuf::from)
+}
+
+/// Classify a raw destination `sockaddr` into the [`DestShape`] the non-IP send
+/// arms key off. Pure function of the already-copied bytes: no child state is
+/// re-read, so the shape cannot change after the decision.
+///
+/// The distinction that matters is the destination's ADDRESS FAMILY, not just
+/// "did a pathname come out of it": an `AF_NETLINK` sockaddr and an abstract
+/// `AF_UNIX` sockaddr both yield no pathname, but they must take opposite
+/// paths — see [`DestShape`].
+pub(crate) fn classify_dest_shape(addr_bytes: &[u8]) -> DestShape {
+    // Below two bytes there is no family to read; fail closed by reporting the
+    // shape the caller refuses.
+    if addr_bytes.len() < 2 {
+        return DestShape::UnixNoPath;
+    }
+    let family = u16::from_ne_bytes([addr_bytes[0], addr_bytes[1]]);
+    if family != libc::AF_UNIX as u16 {
+        return DestShape::NotUnix;
+    }
+    match named_unix_socket_path(addr_bytes) {
+        Some(path) => DestShape::UnixNamed(path),
+        None => DestShape::UnixNoPath,
+    }
 }
 
 /// `struct msghdr` size on LP64 Linux (x86_64 / aarch64 / riscv64): four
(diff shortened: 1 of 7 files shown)


congwang-mk (changes requested):
The TOCTOU diagnosis is correct, and the fail-closed non-unix arm plus the pure `classify_send_path` helper are the right shape. Two problems in the `UnixOnBehalf` arm block this as written:

**1. On-behalf abstract sends bypass Landlock abstract-socket scoping.** The `_` arm receives abstract addresses (`named_unix_socket_path` returns `None` for them; its comment even says "Landlock scope handles it"). `LANDLOCK_SCOPE_ABSTRACT_UNIX_SOCKET` is enforced against the credentials of the process doing the `sendto`. Pre-PR, Continue ran the child's own syscall under the child's Landlock domain. Post-PR the unconfined supervisor performs the send, so enabling an IP destination policy deterministically disables abstract-socket scoping for unconnected sendto: the child can message any abstract socket on the host. That trades a racy IP-policy bypass for a deterministic Landlock-boundary bypass (same shape as #27).

Fix: fail closed for abstract destinations (`EAFNOSUPPORT`), matching what `send_msghdr_on_behalf` already does and what #130 aligns with anyway.

**2. Named paths (fs gate off) resolve in the supervisor's context.** The arm passes the raw child `addr_bytes` to `resolve_send`, so `sun_path` resolves against the supervisor's cwd/root instead of the child's. A relative `sun_path` silently targets the wrong path, and a chroot child's absolute path would resolve host-side. Every other named-unix on-behalf path resolves via `/proc/<pid>/root` (`resolve_named_unix_target`) for exactly this reason.

Fix: pin the inode via a `/proc/<pid>/root` resolution variant (grant check skipped) and send to `/proc/self/fd/<pin>`; rejecting relative `sun_path` with an errno is fine, resolving it against the wrong cwd is not.

Both fixes are user-visible behavior (abstract datagrams under a destination policy go from racy-working to `EAFNOSUPPORT`), so please note them in the behavior-scope section. No family carve-out needed for AF_VSOCK/AF_PACKET; fail-closed there is fine as flagged.

Also worth adding: a happy-path integration test that a named unix datagram still sends under a `net_allow` policy; that is the path both issues live on, and it would have caught the relative-path regression.

dzerik:
Both fixes are in, plus a third one that the first attempt at (1) turned out to need.

**(1) Abstract sends fail closed.** `network/verdict.rs:181` — an AF_UNIX destination with no pathname (abstract, unnamed, non-UTF-8, or a buffer too short to carry a family) maps to `SendPath::Reject`, which the handler turns into `EAFNOSUPPORT`. Verified end to end rather than by reading the diff: a host process binds an abstract name *outside* the sandbox, and the sandboxed child `sendto`s it under a destination policy. On the pre-fix revision the host listener receives the payload; on the tip it receives nothing and the child sees errno 97. That is `test_abstract_unix_dgram_sendto_refused_under_destination_policy` in `tests/integration/test_landlock.rs`.

**(2) Named paths resolve in the child's root view.** `network/unix.rs:37` — `child_root_path` is now the single builder for `/proc/<pid>/root<sun_path>`, and it refuses a non-absolute `sun_path`. Both consumers go through it: the pre-existing gated arm (`resolve_named_unix_target`) and the new fall-through (`pin_child_unix_target`, `unix.rs:53` — `O_PATH` pin, then send on `/proc/self/fd/<pin>`). A relative `sun_path` ends in the same `ECONNREFUSED` as before, so this adds no new user-visible failure mode.

**(3) Classification had to move to the destination's address family.** Worth flagging, because the obvious form of fix (1) is wrong. "AF_UNIX with no pathname" and "a non-AF_UNIX sockaddr on a unix socket" are different shapes, and collapsing both into the fail-closed arm takes out sandlock's own NETLINK_ROUTE virtualization: its child-side fd is an `AF_UNIX SOCK_SEQPACKET` socketpair (`netlink/handlers.rs`), so a `sockaddr_nl` legitimately arrives on a unix socket. I shipped that collapse first and measured the damage — `socket.if_nameindex()` inside the cage returns `['lo']` on main and `OSError(97)` with the collapse.

So `network/materialize.rs` now derives a `DestShape` from the family first (`UnixNamed` / `UnixNoPath` / `NotUnix`), and `verdict.rs:180-182` maps it: named → gate arm, pathless → `Reject`, non-unix → `RawDestOnBehalf`, which sends on the already-pinned fd with the child's address bytes verbatim. That arm is TOCTOU-safe for a different reason than the named one: the fd is pinned before its (stable) `SO_DOMAIN` is read, and the virtualized netlink socket is a connected SEQPACKET, which ignores `msg_name`. Regression test: `if_nameindex_works_under_destination_policy` in `tests/integration/test_netlink_virt.rs`.

**Behavior scope** is updated for both user-visible changes. One addition you may not expect: `has_net_destination_policy` is also true when a `policy_fn` is set, so `sandlock learn` puts every run into this arm — its observation mode is in scope for the abstract change. Flagging it because that landed after this branch was cut.

**On the happy-path test** — added as `test_named_unix_dgram_sendto_delivers_under_destination_policy`, but I have to correct the premise: it would *not* have caught the relative-path regression, and it exercises neither fix. The sandbox it needs in order to send has fs grants, so `has_unix_fs_gate` is true and the send takes the pre-existing named-unix gate arm, not the fall-through this PR changes. Verified by mutation: gutting `sendto_pinned_unix_on_behalf`, stripping the `/proc/<pid>/root` prefix, and turning the fail-closed arm into `Continue` each leave that test — and the whole suite — green.

What the test does carry is that a named unix datagram is still *delivered* while a destination policy is active; it does not pin down which arm delivers it (mutating the gate arm so the send falls through to `sendto_pinned_unix_on_behalf`, which also pins and delivers, keeps it green too). The properties you asked about are carried by the two tests named above and by the `child_root_path` unit tests. That is written in the test's doc comment rather than left for a reader to rediscover.

**One coverage gap, disclosed rather than papered over.** `sendto_pinned_unix_on_behalf` — the named-unix fall-through — has no test, because I could not construct a configuration that reaches it. It requires `has_unix_fs_gate == false`, i.e. zero fs grants, and then Landlock denies `execve` (measured: `execvp '/bin/true': Permission denied`), while `Sandbox::fork` installs a deny-only filter with no notification supervisor. The arm is kept for the same reason its siblings are, and the reachability argument sits next to it in the source.

Suites on the tip: 297 integration / 529 lib, no failures (main: 294 / 517).

PR #157, opened 2026-07-22
build wheels for py 3.14 and runs tests on 3.13 and 3.14 as well

apollo13:
Added wheel build for 3.14 and enabled tests for 3.13 and 3.14 as well. Wonder if it would make sense to cut down tests to min/max python versions to reduce resource usage?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,12 +116,12 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.22
+        uses: pypa/cibuildwheel@v4.1.1
         with:
           package-dir: python
           output-dir: wheelhouse
         env:
-          CIBW_BUILD: "cp310-manylinux* cp311-manylinux* cp312-manylinux* cp313-manylinux*"
+          CIBW_BUILD: "cp310-manylinux* cp311-manylinux* cp312-manylinux* cp313-manylinux* cp314-manylinux*"
           CIBW_ARCHS_LINUX: ${{ matrix.arch }}
           CIBW_BEFORE_ALL_LINUX: >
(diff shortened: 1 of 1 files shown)


congwang-mk:
Thanks for the PR!

I am a bit hesitated to raise the minimum Python version, since we don't need additional effort to support 3.8, I'd prefer to keep it. Do you see any benefit to raise to 3.10?

For the CI matrix, is there a resource usage issue? It only runs on github, locally we only run it against the local version of Python.

apollo13:
> I am a bit hesitated to raise the minimum Python version, since we don't need additional effort to support 3.8, I'd prefer to keep it. Do you see any benefit to raise to 3.10?

Well I usually argue that if people cannot upgrade to a supported Python version they can stay on older versions of the libraries as well. Happy to remove the commit though. That said since wheels aren't published for 3.8/3.9 it would require rust etc anyways, so it is not easy to get running there at all.

> For the CI matrix, is there a resource usage issue? It only runs on github, locally we only run it against the local version of Python.

There is no resource usage issue per se for you now. But even if the resources you are using are free they still need to be bought by someone and contribute to an overall higher power consumption etc. Hence my question whether you'd get roughly the same confidence level if only testing for 3.10 and 3.14 (I don't know sandlock well enough to know whether python point releases behave differently enough to break it every other release).

congwang-mk:
Two changes I request:

1. Bump cibuildwheel: v2.22 predates 3.14 support, so the new `cp314-manylinux*` selector matches nothing and no 3.14 wheel gets built. Please bump to a 3.14-capable version.
2. Please drop the `requires-python` bump. I'd like to keep 3.8/3.9 installable from sdist. We can discuss raising the minimum separately.

apollo13:
Ok, updated to the latest cibuildwheel, let's see what the CI says or if the action needs further adjustments.

congwang-mk:
Thanks, @apollo13

PR #162, opened 2026-07-25
cow: subtree-aware whiteouts with a durable deletion log and type-aware copy-up

### LINKED ISSUE #158: COW copy-up of a FIFO hangs the supervisor indefinitely
`execute_copy` opens the lower file with `O_RDONLY` and no file-type check (`cow/seccomp.rs:508-514`). Opening a FIFO that way blocks until a writer appears, so a child that touches a FIFO under the workdir wedges the supervisor thread performing the copy-up. The supervisor is not confined by the child's policy and nothing times the open out, so the hang is indefinite.

The project already treats this exact hazard as known in the Landlock path-rule code — `test_path_rule_on_fifo_does_not_block` states it plainly: "Opening a FIFO with `O_RDONLY` blocks until a writer appears. `add_path_rule` must reference the path with `O_PATH` so a read rule on a FIFO does not hang child setup." The COW copy-up path has no equivalent guard.

Reproducer (a unit probe against the branch API, so the wedge is visible without a cage):

```rust
let fifo = workdir.path().join("pipe");
let c = std::ffi::CString::new(fifo.to_str().unwrap()).unwrap();
assert_eq!(unsafe { libc::mkfifo(c.as_ptr(), 0o644) }, 0);

let (tx, rx) = std::sync::mpsc::channel();
std::thread::spawn(move || {
    let mut b = SeccompCowBranch::create(&wd, Some(&sd), 0).unwrap();
    let _ = tx.send(format!("{:?}", b.ensure_cow_copy("pipe").map(|_| "Ok")));
});
match rx.recv_timeout(std::time::Duration::from_secs(5)) {
    Ok(v) => println!("returned: {v}"),
    Err(_) => println!("*** HUNG (no return in 5s) ***"),
}
```

Observed: `*** HUNG (no return in 5s) ***`. The thread never returns.

From inside the sandbox this is one line of unprivileged code — write to a FIFO that exists in the workdir, or create one first if `mknod`/`mkfifo` is reachable under the policy. Severity is availability only: the copy is confined to the workdir root (`openat2_in_root`), so this is a wedge, not an escape.

Where the fix belongs is `execute_copy`, before the source open — the destination side is fine. Either stat the source and require `S_IFREG`/`S_IFLNK`, or open `O_RDONLY | O_NONBLOCK` and reject anything that is not a regular file after the fact. The `EACCES`/`ENOENT` arm two lines below already establishes the pattern for "give the child an empty COW file and move on", which looks like the right disposition for a non-regular source as well — a FIFO's contents are not meaningful to copy up.

Found while writing COW merge tests for #148; unrelated to that PR's subject, hence a separate report. Reproduced on `main` at 44b1c65.

### LINKED ISSUE #159: Whiteout is not recursive: a deleted directory's contents stay readable and can be republished by commit()
`is_deleted` is an exact string match on the whiteout set (`cow/seccomp.rs:384-386`), so a whiteout covers the path it was recorded for and nothing beneath it. After a directory is marked deleted, its children are still resolved against the workdir: they stay readable, they stay listable, and a write-open of one copies the deleted file's bytes into the upper, where `commit()` republishes them into the workdir.

Reproducer:

```rust
fs::create_dir(workdir.path().join("d")).unwrap();
fs::write(workdir.path().join("d/secret.txt"), "SECRET").unwrap();

let mut b = SeccompCowBranch::create(workdir.path(), Some(storage.path()), 0).unwrap();
b.mark_deleted("d");

b.is_deleted("d")                    // true
b.is_deleted("d/secret.txt")         // false
b.list_merged_dir("d")               // ["secret.txt"]
b.handle_open("d/secret.txt", 0)     // Ok(None)  -> resolves to the workdir path
```

`Ok(None)` from `handle_open` means the syscall is not intercepted and the child reads the workdir file directly, so the contents of a directory the child deleted remain fully accessible under it.

This diverges from the overlay semantics the branch models elsewhere: `handle_open` on a whiteouted path itself is careful to return `BranchError::Deleted` rather than fall through to the lower file, and its comment explains why ("the lower file still physically exists with its pre-delete bytes"). The same reasoning applies one level down, but the lookup never gets there.

Two consequences worth separating:

* **Visibility** — a deleted directory's contents are readable and listable. The child deleted them itself, so this is a correctness divergence from `rm -rf` semantics rather than a boundary crossing.
* **Integrity** — this is the one that surprised me. A write-open under the deleted directory copies the file up, and the merge then writes it back into the workdir, so a file the transaction deleted reappears in the committed result. That is a silently wrong commit, not just a stale read.

The narrow fix is to make the whiteout check prefix-aware (a path is deleted if it or any ancestor is in the set), with the usual care that `d` must not shadow `d2`. The broader question is whether `mark_deleted` on a directory should record an opaque marker instead, which is what overlayfs does and what a future recovery sweep would need in order to replay a deletion correctly.

Found while writing COW merge tests for #148; unrelated to that PR's subject, hence a separate report. Reproduced on `main` at 44b1c65.

### LINKED ISSUE #160: Renaming a lower-only directory silently destroys its contents
Renaming a directory that exists only in the lower layer reports success and destroys everything inside it.

`handle_rename` (`cow/seccomp.rs:804`) starts with `ensure_cow_copy(&old_rel)`. For a directory that produces an empty directory in the upper — the copy-up is not recursive — and the following `renameat_in_root(&self.upper, &old_rel, &new_rel)` then moves that empty directory to the new name. The source still exists in the workdir, so `mark_deleted(&old_rel)` records a whiteout for it, and `commit()` applies the deletion. The result is a new empty directory and a deleted old one, with every file that was under it gone.

Reproducer:

```rust
let wd = workdir.path().canonicalize().unwrap();
fs::create_dir(wd.join("d")).unwrap();
fs::write(wd.join("d/inner.txt"), "PRECIOUS").unwrap();

let mut b = SeccompCowBranch::create(&wd, Some(storage.path()), 0).unwrap();
b.handle_rename(&format!("{}/d", wd.display()), &format!("{}/d2", wd.display()));
b.commit();
```

Observed:

```
rename          = Ok(true)
changes         = [Change { kind: Deleted, path: "d" }]
commit          = Ok(())
d exists        = false
d2 exists       = true
d2/inner.txt    = false   <-- destroyed
```

Three things make this worse than a plain bug:

* it reports success at every step — the child gets a successful `rename(2)`, `commit()` returns `Ok(())`, and nothing surfaces an error;
* `changes()` reports only `Deleted "d"`, so neither the dry-run output nor the recovery report mentions the files that were lost;
* the whiteout is what does the destroying, so the data is gone from the workdir at commit time even though the upper never held it.

The same shape reaches `handle_link` (`cow/seccomp.rs:872`), which also calls `ensure_cow_copy` on a path that may be a directory.

The fix belongs in the copy-up: either make `prepare_copy`/`ensure_cow_copy` recursive for directories, or have the directory case refuse (return `Ok(false)` and let the kernel perform the rename on the workdir directly, which is what happens today when `safe_rel` rejects the path and is the behaviour that keeps the contents). The second is much cheaper and loses only the ability to stage a directory rename in the branch.

Reproduced on `main` at 44b1c65; found while writing COW merge tests for #148, unrelated to that PR's subject.

Note on the path form, since it cost me a false negative first time round: `safe_rel` uses `pathdiff::diff_paths(path, workdir)`, so these handlers expect the absolute path as the child sees it. Passing a relative `"d"` makes `safe_rel` return `None` and the handler answers `Ok(false)` — the operation falls through to the kernel and no data is lost, which looks like the bug is absent.

### LINKED ISSUE #161: rmdir on a non-empty directory succeeds where the kernel gives ENOTEMPTY, and commit() then destroys the contents
`handle_unlink(path, is_dir = true)` accepts a directory that is not empty and answers `Ok(true)`, where `rmdir(2)` would return `ENOTEMPTY`. The whiteout is recorded, and `commit()` then removes the directory and everything under it.

Reproducer:

```rust
let wd = workdir.path().canonicalize().unwrap();
fs::create_dir(wd.join("d")).unwrap();
fs::write(wd.join("d/inner.txt"), "DATA").unwrap();

let mut b = SeccompCowBranch::create(&wd, Some(storage.path()), 0).unwrap();
b.handle_unlink(&format!("{}/d", wd.display()), true);   // Ok(true)
b.commit();                                              // Ok(())
```

Observed: `rmdir = Ok(true)`, `commit = Ok(())`, and `d/inner.txt` is gone afterwards.

Two separate problems sit on top of each other:

* **Divergence from the syscall.** A child that calls `rmdir()` on a directory it believes to be empty gets success instead of `ENOTEMPTY`. Code that relies on that errno to detect leftovers — a build script cleaning a tree, a test asserting a directory was drained — takes the wrong branch, and inside the sandbox it cannot tell.
* **Destructive follow-through.** Because the whiteout is path-based, `commit()` applies it with `remove_dir_all_in_root`, so contents the child never asked to delete are destroyed. `changes()` reports a single `Deleted "d"`, so neither the dry run nor the recovery report names them.

The emptiness check has to happen against the merged view rather than either layer alone: a directory can be empty in the upper while the lower still has entries, and vice versa when the upper holds whiteouts. `list_merged_dir` (`cow/seccomp.rs:1008`) already computes exactly that view, so the check is available — the rmdir path just does not consult it.

Related, and probably wanting one fix between them: #159 (a whiteout is not recursive, so a deleted directory's contents stay readable and can be republished). Both come from the whiteout being a plain path string with no notion of a subtree.

Reproduced on `main` at 44b1c65; found while writing COW merge tests for #148, unrelated to that PR's subject. As in #160, these handlers take the absolute path as the child sees it — a relative argument makes `safe_rel` return `None` and the handler answers `Ok(false)`, which hides the behaviour.

congwang-mk:
Redesign of the seccomp COW branch's deletion model. The flat exact-match `HashSet<String>` (inherited from the Python-era `cowfs/_branch.py`, carried verbatim into the Rust rewrite) is replaced by a subtree-aware whiteout set, and copy-up becomes file-type-aware.

Fixes #158
Fixes #159
Fixes #160
Fixes #161

## The model

A new `DeletionSet` (`cow/deletions.rs`) owns whiteout semantics:

- `covers(rel)` matches per path component: a whiteout on `d` hides `d/x` and `d/x/y`, never `d2`.
- Entries are never removed. A path re-created after deletion is exposed by existing in the upper (upper presence shadows the whiteout), which makes `rmdir d; mkdir d` an opaque directory for free: the new `d` is visible, its old contents stay hidden.
- Every insertion is appended to `deleted.log` beside `upper/` and fdatasynced, so the deletion half of a change set is now self-describing on disk, like the upper already was. RAM stays the in-process source of truth; the log is the durability record a future recovery sweep (RFC #65 Phase 3, PR #148's preserved branches) can replay.

All merged-view queries (`is_deleted`, open/stat/readlink, `list_merged_dir`, `changes()`, `commit()`) route through it, instead of each handler re-implementing its own slice of overlay semantics.

## Per issue

- **#159**: whiteouts now cover the subtree; reads and listings under a deleted directory give ENOENT, and a write-open under it starts fresh instead of copying the deleted bytes up for `commit()` to republish. Also fixes a latent bug of the same family found during the work: `O_CREAT` on a whiteouted file used to resurrect the pre-delete lower bytes.
- **#161**: rmdir checks emptiness against the merged view and gives ENOTEMPTY, as the kernel would; the whiteout can no longer delete a subtree the child never emptied.
- **#160**: renaming a directory stages the whole tree in the upper via a recursive, confined, quota-accounted copy-up; contents survive the commit, and already-whiteouted children do not reappear under the new name. A whiteouted rename/link source now gives ENOENT instead of falling through to the kernel against the lower entry.
- **#158**: copy-up dispatches on file type and never opens a non-regular source. FIFOs, sockets and device nodes are virtualized as empty regular stubs in the upper, created without reading them, per the disposition suggested in the issue. This keeps the COW virtualization contract (`> /dev/null` inside a learn-mode tree needs no real permission on `/dev` and never touches the device) and removes the supervisor wedge. `execute_copy` additionally guards with `O_NONBLOCK` + fstat against the entry changing type between the classifying lstat and the open.

## The invariant

One test pins the property all four issues violated: the merged view the child observed during the run is exactly what `commit()` publishes. Plus each issue's own reproducer as a unit test, and three end-to-end regressions driving a real sandboxed child (`rm -r`, `mv dir`, `rmdir` non-empty).

## Compatibility

No public API, CLI, or SDK signature changes. Child-visible behavior changes only toward kernel semantics (ENOTEMPTY, ENOENT under deleted dirs, dir renames preserving contents). The branch storage dir gains `deleted.log`; the layout is internal.

Suites: 555 lib / 308 integration / 386 Python locally; the two `cli_test` failures on my machine (`test_learn_captures_openat2`, `test_write_collapse_skips_protected`) reproduce identically on main and are unrelated.

### DIFF
--- a/crates/sandlock-core/src/chroot/dispatch.rs
+++ b/crates/sandlock-core/src/chroot/dispatch.rs
@@ -986,8 +986,8 @@ pub(crate) async fn handle_chroot_write(
                 if cow.matches(&old_str) {
                     match cow.handle_rename(&old_str, &new_host.to_string_lossy()) {
                         Ok(true) => return NotifAction::ReturnValue(0),
-                        Err(crate::error::BranchError::QuotaExceeded) => return NotifAction::Errno(libc::ENOSPC),
-                        _ => {}
+                        Err(errno) => return NotifAction::Errno(errno),
+                        Ok(false) => {}
                     }
                 }
             }
--- a/crates/sandlock-core/src/cow/deletions.rs
+++ b/crates/sandlock-core/src/cow/deletions.rs
@@ -1,0 +1,192 @@
+//! Subtree-aware whiteout tracking for the seccomp COW branch.
+//!
+//! A whiteout entered here hides the entered path AND everything beneath it
+//! in the merged view, the way removing a directory hides its contents.
+//! Entries are never removed: a path re-created after deletion is exposed by
+//! existing in the upper layer (upper presence shadows the whiteout), which
+//! is what makes an entry whose directory reappears behave as an opaque
+//! directory. The set therefore only grows, and the on-disk log is
+//! append-only.
+//!
+//! The log exists so the deletion half of a change set survives the process:
+//! the upper directory is self-describing on disk, the RAM-only deletion set
+//! was not (issues #159/#160/#161 all grew from that asymmetry). RAM stays
+//! the in-process source of truth; a log write failure degrades durability,
+//! not correctness, so it is deliberately not fatal.
+
+use std::collections::BTreeSet;
+use std::fs;
+use std::io::{BufRead, Write};
+use std::path::Path;
+
+pub(crate) struct DeletionSet {
+    entries: BTreeSet<String>,
+    log: Option<fs::File>,
+}
+
+fn escape(raw: &str) -> String {
+    let mut out = String::with_capacity(raw.len());
+    for c in raw.chars() {
+        match c {
+            '\\' => out.push_str("\\\\"),
+            '\n' => out.push_str("\\n"),
+            _ => out.push(c),
+        }
+    }
+    out
+}
+
+fn unescape(raw: &str) -> String {
+    let mut out = String::with_capacity(raw.len());
+    let mut it = raw.chars();
+    while let Some(c) = it.next() {
+        if c != '\\' {
+            out.push(c);
+            continue;
+        }
+        match it.next() {
+            Some('n') => out.push('\n'),
+            Some(other) => out.push(other),
+            None => out.push('\\'),
+        }
+    }
+    out
+}
+
+impl DeletionSet {
+    /// Open a fresh set. `log_path` is opened for appending; `None`, or a
+    /// path that cannot be created, gives a RAM-only set.
+    pub fn create(log_path: Option<&Path>) -> Self {
+        let log = log_path.and_then(|p| {
+            fs::OpenOptions::new().create(true).append(true).open(p).ok()
+        });
+        Self { entries: BTreeSet::new(), log }
+    }
+
+    /// Rebuild a set from an existing log and reopen the log for append.
+    /// Unreadable log lines are skipped, not fatal: a torn final line from a
+    /// crash must not hide the deletions before it.
+    // No production caller yet: this is the recovery half of the durability
+    // story, consumed by the preserved-branch sweep (PR #148 / RFC #65).
+    #[allow(dead_code)]
+    pub fn load(log_path: &Path) -> Self {
+        let mut entries = BTreeSet::new();
+        if let Ok(f) = fs::File::open(log_path) {
+            for line in std::io::BufReader::new(f).lines().map_while(Result::ok) {
+                if !line.is_empty() {
+                    entries.insert(unescape(&line));
+                }
+            }
+        }
+        let log = fs::OpenOptions::new().create(true).append(true).open(log_path).ok();
+        Self { entries, log }
+    }
+
+    /// Record `rel` as deleted. Idempotent; the log line and its fdatasync
+    /// happen only on first insertion.
+    pub fn insert(&mut self, rel: &str) {
+        if !self.entries.insert(rel.to_string()) {
+            return;
+        }
+        if let Some(f) = self.log.as_mut() {
+            let _ = writeln!(f, "{}", escape(rel));
+            let _ = f.sync_data();
+        }
+    }
+
+    /// True when `rel` or any ancestor directory of it has been deleted.
+    /// Matching is per path component: "d" covers "d/x" but never "d2".
+    pub fn covers(&self, rel: &str) -> bool {
+        if self.entries.contains(rel) {
+            return true;
+        }
+        let mut end = rel.len();
+        while let Some(i) = rel[..end].rfind('/') {
+            if self.entries.contains(&rel[..i]) {
+                return true;
+            }
+            end = i;
+        }
+        false
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &str> {
+        self.entries.iter().map(String::as_str)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn covers_exact_and_descendants_not_siblings() {
+        let mut s = DeletionSet::create(None);
+        s.insert("d");
+        assert!(s.covers("d"));
+        assert!(s.covers("d/x"));
+        assert!(s.covers("d/x/y"));
+        assert!(!s.covers("d2"));
+        assert!(!s.covers("d2/x"));
+        assert!(!s.covers("e"));
+    }
+
+    #[test]
+    fn covers_nested_entry() {
+        let mut s = DeletionSet::create(None);
+        s.insert("a/b");
+        assert!(!s.covers("a"));
+        assert!(s.covers("a/b/c"));
+        assert!(!s.covers("a/bc"));
+    }
+
+    #[test]
+    fn insert_is_idempotent() {
+        let mut s = DeletionSet::create(None);
+        s.insert("f");
+        s.insert("f");
+        assert_eq!(s.iter().collect::<Vec<_>>(), vec!["f"]);
+    }
+
+    #[test]
+    fn log_roundtrip() {
+        let dir = tempfile::tempdir().unwrap();
+        let log = dir.path().join("deleted.log");
+        {
+            let mut s = DeletionSet::create(Some(&log));
+            s.insert("plain.txt");
+            s.insert("dir/with\nnewline");
+            s.insert("back\\slash");
+        }
+        let replayed = DeletionSet::load(&log);
+        assert_eq!(
+            replayed.iter().collect::<Vec<_>>(),
+            vec!["back\\slash", "dir/with\nnewline", "plain.txt"]
+        );
+        assert!(replayed.covers("dir/with\nnewline/child"));
+    }
+
+    #[test]
+    fn load_appends_not_truncates() {
+        let dir = tempfile::tempdir().unwrap();
+        let log = dir.path().join("deleted.log");
+        {
+            let mut s = DeletionSet::create(Some(&log));
+            s.insert("one");
+        }
+        {
+            let mut s = DeletionSet::load(&log);
+            s.insert("two");
+        }
+        let s = DeletionSet::load(&log);
+        assert_eq!(s.iter().collect::<Vec<_>>(), vec!["one", "two"]);
+    }
+
+    #[test]
+    fn ram_only_when_no_log() {
+        let mut s = DeletionSet::create(None);
+        s.insert("x");
+        assert!(s.covers("x"));
+        assert_eq!(s.iter().count(), 1);
+    }
+}
--- a/crates/sandlock-core/src/cow/deletions.rs
+++ b/crates/sandlock-core/src/cow/deletions.rs
@@ -30,6 +30,10 @@ fn escape(raw: &str) -> String {
         match c {
             '\\' => out.push_str("\\\\"),
             '\n' => out.push_str("\\n"),
+            // BufRead::lines strips a trailing \r on replay; unescaped, a
+            // path ending in \r would round-trip to the \r-less sibling,
+            // hiding that one and resurrecting the deleted path.
+            '\r' => out.push_str("\\r"),
             _ => out.push(c),
         }
     }
@@ -46,6 +50,7 @@ fn unescape(raw: &str) -> String {
         }
         match it.next() {
             Some('n') => out.push('\n'),
+            Some('r') => out.push('\r'),
             Some(other) => out.push(other),
             None => out.push('\\'),
         }
@@ -53,12 +58,26 @@ fn unescape(raw: &str) -> String {
     out
 }
 
+/// Persist the log's directory entry. `insert` only fdatasyncs the file
+/// itself, which does not cover the name in the parent directory: without
+/// this, a power loss after the first deletion can drop the whole log, the
+/// exact crash class it exists for. Best-effort like every other log write.
+fn sync_parent_dir(p: &Path) {
+    if let Some(dir) = p.parent() {
+        if let Ok(d) = fs::File::open(dir) {
+            let _ = d.sync_all();
+        }
+    }
+}
+
 impl DeletionSet {
     /// Open a fresh set. `log_path` is opened for appending; `None`, or a
     /// path that cannot be created, gives a RAM-only set.
     pub fn create(log_path: Option<&Path>) -> Self {
         let log = log_path.and_then(|p| {
-            fs::OpenOptions::new().create(true).append(true).open(p).ok()
+            let f = fs::OpenOptions::new().create(true).append(true).open(p).ok()?;
+            sync_parent_dir(p);
+            Some(f)
         });
         Self { entries: BTreeSet::new(), log }
     }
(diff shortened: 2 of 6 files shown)


dzerik:
Thanks for taking on the whole deletion model — the subtree-aware whiteouts and the merged-view `rmdir`/`ENOTEMPTY` are a real improvement. From the reporter side I can confirm **#161 is fully closed** (opaque dir after `rmdir d; mkdir d`), **#159's core is closed** (reads, listings and `O_CREAT` under a deleted dir all `ENOENT`/start-fresh — the `O_CREAT`-no-resurrect bonus holds), and **#160's reported case is closed** (a renamed dir keeps its children, whiteouted children do not reappear, a whiteouted source gives `ENOENT`). Type-aware copy-up does close the copy-up-time hang.

I re-ran each of the four as an in-cage test against this branch. A few things I would want resolved before it lands.

**Blocker — #158 is only half-closed; the hang is relocated to `commit()`.**

`prepare_copy` (`cow/seccomp.rs:288-306`) virtualizes a FIFO/socket/device as an empty *regular* upper stub but never `mark_deleted`s the surviving lower special file. `commit()` applies deletions first (`943-950` — the FIFO is not in the set), then walks the upper, sees the stub as a regular file, and at `989-995` does `openat2_in_root(workdir, rel, O_WRONLY|O_CREAT|O_TRUNC|O_NOFOLLOW)`. That destination *is* the surviving lower FIFO: `O_NOFOLLOW` does not affect FIFOs and `openat2_in_root` adds no `O_NONBLOCK`, so an `O_WRONLY` open of a reader-less FIFO blocks forever. `commit()` runs in `impl Drop` with the error swallowed, so this is an unrecoverable supervisor hang, not a failed merge.

Repro: lower FIFO `pipe`, child does `echo hi > pipe`, `BranchAction::Commit` — the commit blocks (>25s watchdog, `pipe` still a FIFO on disk). The identical child with `Abort` passes, which isolates it to the commit publish path. The existing FIFO tests (`copy_up_of_fifo`, `write_open_of_fifo`, `chmod_of_fifo`) all stop before `commit()`, so CI stays green.

Fix options: `mark_deleted` the lower path when stubbing a non-regular entry (commit then unlinks-then-creates), unlink a type-mismatched destination before opening, or open the destination `O_NONBLOCK` and handle `ENXIO`.

**Should-fix — a directory rename onto an existing lower-only destination silently merges the two trees (new, reachable via #160).**

`handle_rename` (`seccomp.rs:688-699`) stages the source into the upper with `copy_up_tree` and then `renameat_in_root(upper, old, new)`. When `new` exists only in the lower it is never whiteouted, so the merged view of `new` becomes the union of the renamed tree and the pre-existing lower tree, and the syscall returns success. With lower-only `a/{a1}` and `b/{b1}`, `mv -T a b` returns 0 and commits `b = {a1, b1}` — a state that never existed. POSIX wants `ENOTEMPTY` for a non-empty destination, or an atomic replace.

**Should-fix — #159 gap: the `O_DIRECTORY` open path never consults `is_deleted`/`covers`.**

`handle_open` returns `Ok(None)` for `O_DIRECTORY` unconditionally (`450-452`), and `prepare_open`'s `O_DIRECTORY` branch resolves to the upper only when `upper_dir.is_dir() && !lower_dir.is_dir()` (`519`), otherwise `Skip` → `Continue` → the kernel opens the lower directory. After a normal `rm -rf d` of a lower-only dir, `stat("d")` gives `ENOENT` but `open("d", O_DIRECTORY)` succeeds on the lower fd, and `fstat` on it leaks the deleted directory's inode/nlink/mtime. The blast radius is bounded (getdents is virtualized empty via `list_merged_dir`, and `openat(dirfd, child)` is re-gated through absolute-path reconstruction), but an open succeeding where the merged answer is `ENOENT` breaks the "present as absent" invariant this PR pins elsewhere. Short-circuiting `is_deleted(rel)` before the `O_DIRECTORY` split in both functions closes it.

**Should-fix — `copy_up_tree` failure paths (`seccomp.rs:402-426`).**

`read_dir`/`statat` errors are swallowed as `Ok(())` (`406`, `416`), so a mid-tree `EACCES`/`EIO` truncates the staged copy — and `handle_rename` then `mark_deleted`s the source anyway (`697`), so the untraversed children are silently lost at commit. It also recurses with no depth bound over child-controlled directory depth (a pre-seeded deep lower tree can overflow the supervisor's stack), and a mid-tree `QuotaExceeded` leaves `disk_used` charged and an orphan upper subtree behind with no rollback. Propagating those errors and making the staging all-or-nothing would match the guarantees the rest of the branch gives.

**Nice-to-have — `deleted.log` format and durability (latent while `load()` is `dead_code`).**

`escape()` (`deletions.rs:27-37`) escapes `\\` and `\n` but not `\r`, while `load()` reads via `BufRead::lines()`, which strips a trailing `\r`. A file named `"d/file\r"` round-trips to `"d/file"`, after which `covers()` hides the wrong sibling and the actually-deleted path resurrects. Separately, `create()` never fsyncs the parent `branch_dir` and `insert()` only `sync_data()`s the log file, so a power loss after the first deletion can lose the log's directory entry entirely — the exact crash class the log exists for. Escaping `\r` (or length-delimiting) and one fsync of `branch_dir` after create would close both.

**Coupling with #148 — worth deciding the sequencing before either merges.**

These two rework the same `SeccompCowBranch` commit/changes/deletion surface, and they collide semantically rather than textually:

- `DeletionSet` is append-only with no `remove()`, but #148 mutates its flat set via `deleted.remove()` on re-create in `prepare_copy` and to shrink the retryable remainder in `commit_merge`. Neither is expressible under `covers()` as it stands, so adopting `DeletionSet` in #148 is a redesign (shadow-via-upper-presence plus a separate `applied` set), not a rebase.
- #162's `commit()` is the pre-#148 flat publish path: it classifies the deletion destination with `dest.is_dir()` (`945`), which follows symlinks — the bug #148 fixes with `symlink_metadata()`. It also swallows every deletion error, takes no workdir lock, and writes no interrupted-merge marker. If #162's `commit()` wins the merge, those #148 fixes regress.
- `deletions.rs` describes the log as the durability record #148's preserved branches can replay, but #148's recovery keys entirely on the `PRESERVED` marker (it fills deletions from the marker's `deleted=` lines) and never opens `deleted.log`, so `load()` currently has no producer.

Suggested order: land `DeletionSet` first, then rebuild #148's preserve/marker/remainder logic on a single deletion source of truth — either dropping the log's recovery role or having `preserve()` snapshot `DeletionSet::iter()` into the marker. Happy to drive that on the #148 side.

Only the #158 commit hang is a hard blocker for me; the rest is should-fix/nice-to-have plus the sequencing call.

dzerik:
Pushed the repros so you can run them directly: `dzerik/sandlock` branch `review/pr-162-cow-repros`, based on this PR's head (`c712baf`), test file only.

```
cargo test -p sandlock-core --test integration hunt_ -- --test-threads=1
```

Three fail here — `hunt_158_fifo_commit_hang`, `hunt_159_opendir_whiteouted_dir_leaks_lower`, `hunt_160_rename_onto_existing_dir_merges` — and the rest pass as the reporter-side confirmations that #161, #159's core and #160's reported case are closed.

Two notes on how they are written: the FIFO one drives `run()` + drop on its own OS thread and detects the hang by joining with a deadline, because `commit()` runs in `impl Drop` and a tokio timeout cannot interrupt a blocking drop; and the `O_DIRECTORY` one goes through the static rootfs helper, whose `ls` calls `opendir()` with no pre-`stat`, so it hits that open path exactly rather than being answered by the (correct) stat path. All children use only sh, coreutils and the helper — python3 does not start in every cage environment, which would otherwise mask a result.

congwang-mk:
Thanks for the reproducer and review, @dzerik . I have addressed them. Please take another look.

dzerik:
Thanks — I re-ran the repros against `ae7df2b`.

Closed on my side: the `O_DIRECTORY` gap, the rename-onto-existing-destination merge, the `copy_up_tree` failure paths, and the `deleted.log` `\r` / directory-fsync issues. `hunt_159_opendir_whiteouted_dir_leaks_lower` and `hunt_160_rename_onto_existing_dir_merges` now pass, and the earlier confirmations (#161, #159's core, #160's reported case) still do.

The #158 commit-time hang still reproduces, though: `hunt_158_fifo_commit_hang` blocks for >25s and the child's bytes are never published.

`mark_deleted` is still called only from `handle_unlink` (`cow/seccomp.rs:713`) and `handle_rename` (`:801`, `:804`), never from the non-regular stub path (`~:308-320`). So the lower FIFO survives unwhiteouted, `commit()`'s deletion pass does not remove it, and the upper walk then `O_WRONLY`-opens that surviving FIFO as its destination and blocks on a reader that never comes — in `impl Drop`, with the error swallowed.

Copy-up time is genuinely fixed; it is only the publish path that is left. Whiteouting the lower path when the stub is created (so commit unlinks-then-creates), unlinking a type-mismatched destination before opening it, or opening the destination `O_NONBLOCK` and handling `ENXIO` would each close it.

PR #166, opened 2026-07-25
fix(sandbox): drain the capture pipes while the child runs (#149)

### LINKED ISSUE #149: Sandbox.run() deadlocks when the child writes more than the pipe buffer to stdout/stderr
## Summary

`Sandbox.run()` (the one-shot convenience API) deadlocks and hangs until the timeout when the sandboxed child writes more than one pipe buffer worth of data to a piped `stdout`/`stderr`. Observed hang threshold in practice: as little as **32 KiB** of stdout. Below the buffer it returns instantly; at/above it the run blocks for the full `timeout`, then the child is SIGKILLed and a `timeout` result is returned — the actual output is lost.

## Root cause

`run()` → `wait()` waits for the child to fully exit **before** reading the capture pipes:

```rust
// crates/sandlock-core/src/sandbox.rs (wait)
let exit_status = match self.rt_mut().pidfd.take() {
    Some(pidfd) => wait_child_exit_via_pidfd(pidfd, pid).await,  // waits for EXIT
    None => wait_child_exit_blocking(pid).await,
};
...
let stdout = self.rt_mut()._stdout_read.take().map(sandbox_read_fd_to_end); // read AFTER exit
let stderr = self.rt_mut()._stderr_read.take().map(sandbox_read_fd_to_end);
```

The read end of the stdout/stderr pipe is drained **only after the child exits**. But when the child fills the pipe buffer, its `write(2)` blocks (the buffer is never drained during execution), so the child **never exits**. `wait_child_exit` then blocks until the timeout. Classic pipe-full deadlock — the same failure mode `subprocess` avoids with `communicate()` (concurrent drain).

## Why `run()` specifically is a bug (not just the documented contract)

The low-level `popen()`/`Process.wait()` API *documents* the drain-before-wait contract:

> a piped `stdout`/`stderr` you have not drained can block the child on a full > pipe and hang the wait forever; pass a `timeout` or drain first.

For `popen()` the caller owns the read fd and can drain concurrently, so that is a reasonable contract. But `run()` performs `wait` + read **internally** and never hands the read fd to the caller, so there is **no seam** to drain from. The contract is therefore unsatisfiable for `run()`, which makes any child producing more than a pipe buffer of output a hang. `run()` is documented as
"the common one-shot case" with no size caveat, so callers reasonably expect it to handle arbitrary output sizes like `subprocess.run(..., capture_output=True)`.

## Reproduction

```python
from sandlock import Sandbox

sb = Sandbox(...)  # any policy; stdout defaults to Piped 
res = sb.run(["python3", "-c", "print('x' * 65536)"], timeout=25)
# Expected: res.stdout == b"x"*65536 + b"\n", quick return
# Actual:   blocks ~25s, child SIGKILLed, timeout result, output lost
```

`print('x' * 32767)` returns instantly; `print('x' * 32768)` hangs.

## Suggested fix

Drain the capture pipes **concurrently** with the wait inside `run()`/`wait()`, e.g. spawn Tokio tasks that `read_to_end` the stdout/stderr fds and `join` them alongside `wait_child_exit` (the `communicate()` pattern). This removes the
deadlock for any output size while keeping the current `run()` signature.

An alternative/companion would be to expose an fd-passing variant in the public API (the Rust core already has `create_with_io`) so callers can redirect the child's stdout to a regular file / their own fd and avoid the pipe entirely.

## Impact

Any `run()` caller whose sandboxed workload legitimately produces >~32 KiB of stdout/stderr (large prints, verbose libraries, serialized data) hangs for the full timeout and loses the output. There is currently no way to work around it  through the public Python API on the released version other than moving the capture off the pipe from inside the child.

dzerik:
Fixes #149.

`wait()` read the capture pipes only after awaiting child exit. A child that writes more than one pipe buffer — 64 KiB by default — blocks in `write()` once the pipe fills, so it never reaches the exit `wait()` is waiting for: the run hangs until the caller's timeout, forever if there is none, and the output is lost. On current main, 64 KiB returns in 17 ms and 1 MiB never returns.

The drains now start before the exit wait and are joined after it. They are ordinary async tasks (`tokio::net::unix::pipe::Receiver`), so no blocking-pool threads and no new OS threads are used — which also keeps `run()` usable where thread creation is refused (#47). `O_NONBLOCK` is set on the supervisor's read end only; the child writes the *write* end, a separate open file description, so its writes stay blocking and back-pressure is preserved.

The handles are parked on the `Sandbox` rather than on the `wait()` future, so cancelling `wait()` no longer loses the capture: a later `wait()` joins the drains and returns the complete output. Before this change a retry returned only what had fit in one pipe buffer, and the `RuntimeState::Stopped` early return reported nothing at all. `Drop` aborts the drains.

Streams taken through `Process::take_stdout` / `take_stderr` are unaffected and remain the caller's to drain.

## Scope

Covers every path that captures through `wait()`: `run`, `run_interactive`, `run_with_handlers`, `dry_run`, `Process::wait`.

Not covered: `Pipeline::run` / `Gather::run`, which read their own pipes after the stage-wait loop, and `fork` / `reduce`, which blocks in `sandbox_read_exact` before the reducer drains the clone pipes. Same defect, deliberately unchanged here — happy to follow up separately if you want it in one go.

## Properties worth stating

**EOF liveness is unchanged.** The drains end at EOF, which requires every holder of the write end to close it, so a descendant that inherited fd 1/2 and outlives the child still keeps `wait()` blocked. That is exactly the previous behaviour — the old post-exit `read_to_end` blocked on the same condition, inline and uncancellable.

**Memory trade.** Capture is now bounded by the workload rather than by the pipe: `wait()` holds the whole stream in memory, where the pipe buffer previously stopped it at 64 KiB by hanging. Callers running output-unbounded workloads should take the stream and cap it themselves. A `capture_limit` builder option would be a reasonable follow-up; note that a capped reader must still drain to EOF or it reintroduces this issue.

## Tests

Three over-buffer cases — stdout, stderr, and both at once. The combined one is the discriminating case: a *sequential* drain still deadlocks on it, because stdout only reaches EOF when the child exits and the child is blocked writing stderr.

Two more for the properties above: cancel-then-retry must still return the capture, and a runtime with `max_blocking_threads(2)` must not deadlock a second capture-mode run while a zero-output sandbox is alive.

Each fails without its part of the change, and all are timeout-bounded so a regression fails the test instead of hanging CI.

### DIFF
--- a/crates/sandlock-core/src/sandbox.rs
+++ b/crates/sandlock-core/src/sandbox.rs
@@ -255,6 +255,11 @@ struct Runtime {
     loadavg_handle: Option<JoinHandle<()>>,
     _stdout_read: Option<std::os::fd::OwnedFd>,
     _stderr_read: Option<std::os::fd::OwnedFd>,
+    // Tasks draining the capture pipes above while the child runs (see `wait`).
+    // They belong to the runtime, not to the `wait()` that started them, so a
+    // cancelled `wait()` does not take the output with it.
+    stdout_drain: Option<JoinHandle<Vec<u8>>>,
+    stderr_drain: Option<JoinHandle<Vec<u8>>>,
     // Parent-held write end of a piped stdin (popen). The caller takes it via
     // `Process::take_stdin`; closing it signals EOF to the child.
     _stdin_write: Option<std::os::fd::OwnedFd>,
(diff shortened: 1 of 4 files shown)


congwang-mk:
Nice fix overall. One gap: `collect_pipe_drains` takes the handles out of the runtime before awaiting them, so cancelling `wait()` during the final join drops the `JoinHandle`s and the retry returns no output. That's reachable in the documented slow case (a descendant holding the write end keeps the join blocked past the caller's timeout), and it contradicts the cancel-safety promise in the `wait()` docs.

Suggest awaiting through `as_mut()` and clearing the slot only on completion, so a cancelled join leaves the handle parked for the next `wait()`:

```rust
let stdout = match self.rt_mut().stdout_drain.as_mut() {
    Some(h) => {
        let buf = h.await.unwrap_or_default();
        self.rt_mut().stdout_drain = None;
        Some(buf)
    }
    None => None,
};
```

dzerik:
Good catch — fixed in `78bb08b`. `join_parked_drain` now awaits through the slot and clears it only once the await has produced a value, so a cancelled join leaves the handle where the next `wait()` finds it.

Your snippet doesn't quite borrowck as written: the temporary `&mut Runtime` from `self.rt_mut()` in the match scrutinee lives for the whole `match`, so the second `self.rt_mut()` inside the arm conflicts. It landed as a small free helper taking `&mut Option<JoinHandle<Vec<u8>>>` instead — same shape, and one place covering both streams and the `RuntimeState::Stopped` early return.

Pinned both ways. A unit test asserts the contract deterministically: cancel the join, the handle is still parked, then complete it and the bytes survive. An integration test drives it end to end, where a background subshell holds the write end past the child's exit. Reaching that state end to end is racy, so the integration test retries and fails if no attempt ever blocks the join, rather than passing vacuously.

One thing I hit while building that test, unrelated to this PR and worth its own look: `wait()` aborts `notif_handle` as soon as the top-level child is reaped, so a surviving descendant's in-flight syscalls start failing with `ENOSYS`. That is what makes the survivor flaky here. Happy to file it separately if it isn't already known.

dzerik:
Found one more of the same family while re-reviewing this, fixed in `d43de75`.

The slots parked the *task*, but a finished join handed its bytes straight back to the `wait()` future and cleared the slot. The two streams are joined one after the other, so the second join is a suspension point for a capture the first has already read in full — a caller's timeout landing there destroyed it, with nothing left on the runtime to recover it from. The narrow window your review closed was the same shape one step earlier, so this is the last place the result could still escape the runtime.

Each slot now holds either the task still reading or the bytes it finished with, and `collect_pipe_drains` finishes both streams before taking either, so no cancellation point sits between a completed capture and its hand-off.

Pinned at both levels, each failing without its half of the change: two unit tests on the slot contract (a pending join stays parked, a finished one parks its bytes, only an explicit take hands them over), and an integration test driving the real collector, where the survivor closes stdout but keeps stderr so the cancellation lands exactly between the two joins.

Issue #164, opened 2026-07-25
two problems when run in a baremetal or a devcontainer


nnWhisperer:
Hello,
I'm using ubuntu 24.04, on both server and the docker devcontainer.
The problem occurs in both the baremetal server and the docker container.

Sandlock partially works. Below are my findings for a 7.0.0-28-generic kernel:
The command used to start an interactive shell here in this experiment is:
`sandlock run -i -r /usr -r /lib -r /lib64 -r /bin -r /etc -w /tmp -- /bin/sh`
1) while it is working, `sandlock list` from an alternative terminal says `No running sandboxes.`.
In the shell, it is possible to do `ls /tmp` and it prints correctly, and `ls .` gives permission errors, expectedly as the current location wasn't shared.

When running python:
```python
from sandlock import Sandbox, confine

sandbox = Sandbox(
    fs_writable=["/tmp/sandbox"],
    fs_readable=["/usr", "/lib", "/etc"],
    max_memory="256M",
    max_processes=10,
    clean_env=True,
)

# Run a command (with optional timeout in seconds)
result = sandbox.run(["python3", "-c", "print('hello')"], timeout=30)
```
2) stops at the instant result line is sent to python repl, saying `[1]+  Stopped                 python`.

The 2nd failure step must be doing an effect like sending a ctrl+z to the python process and freezing it. When I say `fg`, python process continues:
```
$ fg
python
>>> result
Result(success=False, exit_code=-1, error='sandlock_create failed', reason=None, signal=-1)
>>> sandbox
[...]

PR #169, opened 2026-07-27
issue #164 test: cover sandbox listing and REPL terminal ownership

nnWhisperer:
I'm sending the test case so that it will be in the CI/CD and will never be broken again. This test case is so fundamental that whose brokenness casts a shadow on the project, which I wouldn't want to. This is a nice project.


congwang-mk:
Thanks for the PR. I ran your Python test locally, it passed with PR #168. (The other one is supposed to be solved with PR #153)

PR #171, opened 2026-07-28
Control socket follow-ups: inspect rename, collision hard-fail, API narrowing

congwang-mk:
Follow-ups to #153 from its final review round.

### `sandlock config` renamed to `sandlock inspect`

"config" universally reads as get/set of the tool's own configuration (git config, npm config) and collides conceptually with sandlock's profile management, while the container ecosystem convention for "show me everything about this running instance" is "inspect" (docker, podman). The new name also leaves room to later include runtime state (ports, uptime, pids) alongside policy. Hard break with no alias, done now while pre-1.0 makes it cheap. The control-socket wire verb stays `config` since it specifically fetches the effective policy.

### `no_supervisor` spawn hard-fails on a live name collision

The `no_supervisor` path warned and continued on `AlreadyExists`, leaving a second sandbox running invisible to `sandlock ps`, which is exactly the outcome the supervisor path already hard-fails to prevent. It now mirrors that arm; other runtime-dir errors (e.g. restricted `/dev/shm` in nested sandboxes) still warn and continue.

The runtime-dir setup also moves after the notif-fd read. An error return from `do_spawn` relies on `Drop`'s `killpg` to reap the forked child, which only works once the child has `setpgid`'d into its own group; the notif-fd write is the child's setup-complete signal. Returning before it races `killpg` against `setpgid`: `killpg` loses with ESRCH, and `Drop` then deadlocks its blocking `waitpid` against a child parked on the ready pipe whose write end `Drop` itself still holds. Found the hard way: the first version of this fix returned inside that window and hung reliably. A comment now pins the invariant, and `test_control_name_collision_no_supervisor` covers the path.

### `control` API narrowed to its external consumers

Everything in the control module was public, but `setup_runtime_dir`, `setup_runtime_dir_no_socket`, and `spawn_control_loop` are lifecycle plumbing only `sandbox.rs` may call; invoking them from outside would corrupt live sandbox state. They and `runtime_dir_uid` are now `pub(crate)`. What stays public each has a real consumer: the client API (`list_live_sandboxes`, `send_control_request`, `ControlResponse`), the on-disk layout helpers (`sandbox_dir`, `pid_path`, `sock_path`), and `cleanup_runtime_dir` (needed by `sandlock kill` because SIGKILL skips the supervisor's `Drop`). The crate root now re-exports only `format_net_rule` from the profile module; the reverse serializers stay reachable under `profile::` and `format_http_rule` becomes `pub(crate)`.

## Test plan

- Control integration tests: 21/21 pass (includes the new collision test, which deadlocked before the ordering fix)
- CLI integration tests: 31/31 pass
- Full release workspace build clean (covers sandlock-oci and sandlock-ffi consumers)

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,8 +217,8 @@ sandlock run --name web.local --port-remap --net-allow-bind 8080 -r /usr -r /lib
 sandlock ps
 
 # Show effective policy for a running sandbox (JSON or TOML)
-sandlock config api.local
-sandlock config api.local --toml
+sandlock inspect api.local
+sandlock inspect api.local --toml
 
 # Kill a running sandbox by name
 sandlock kill web.local
@@ -770,7 +770,7 @@ NAME                                  PID        UPTIME  CMD
 api.local                           12345            5m  python3 server.py
 web.local                           12346            3m  python3 server.py
 
-$ sandlock config api.local --toml | head -10
+$ sandlock inspect api.local --toml | head -10
 [config]
 http_inject_ca = []
 
--- a/crates/sandlock-cli/src/main.rs
+++ b/crates/sandlock-cli/src/main.rs
@@ -23,7 +23,7 @@ enum Command {
     /// List all running sandboxes
     Ps,
     /// Show effective policy for a running sandbox
-    Config {
+    Inspect {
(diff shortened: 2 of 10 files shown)


congwang-mk:

PR #172, opened 2026-07-28
Fix the riscv64 test sweep: hermetic collapse tests, unique sandbox names

### LINKED ISSUE #170: A new failing test on RISCV
```
failures:

---- test_learn_then_run_collapse stdout ----

thread 'test_learn_then_run_collapse' (31781) panicked at crates/sandlock-cli/tests/learn_integration.rs:192:5:
run failed with collapsed profile: cat: /usr/bin/true: Permission denied
hint: if a denial caused this, extend the profile with:
  sandlock learn --merge /tmp/.tmpU8ntZM -- cat /usr/bin/true

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_learn_then_run_collapse

test result: FAILED. 10 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.37s
```

This is probably because on RISCV, the following commands are symlinks:

```
$ ls -lh /usr/bin/cat
lrwxrwxrwx 1 root root 30 Mar 30 09:50 /usr/bin/cat -> ../lib/cargo/bin/coreutils/cat
$ ls -lh /usr/bin/ls
lrwxrwxrwx 1 root root 29 Mar 30 09:50 /usr/bin/ls -> ../lib/cargo/bin/coreutils/ls
$ ls -lh /usr/bin/env
lrwxrwxrwx 1 root root 30 Mar 30 09:50 /usr/bin/env -> ../lib/cargo/bin/coreutils/env
```

congwang-mk:
## Summary

Fixes for every failure from the riscv64 (rust-coreutils, uutils multicall) test sweep. One is a real product bug the sweep exposed; the rest are test portability issues.

### 1. Product bug: internal fixed sandbox names are a global mutex

Since the control-socket work made every sandbox name claim a runtime dir under `/dev/shm/sandlock-$UID/` with a hard failure on live collision, every internally assigned fixed name became a process-wide (in fact UID-wide) mutex: two concurrent pipelines collide on `pipeline-stage-0`, transactions on `txn-stage-0`, gathers on `gather-consumer`, single-stage runs on `stage`. A parallel test run surfaces this immediately (36 core integration tests fail, reproduced on x86_64 with default test parallelism), but any embedder running two pipelines at once hits the same wall.

Fix: factor the auto-name generator's `<pid>-<counter>` suffix into `unique_instance_id()` and stamp it into every internally assigned name. One run id is shared across a pipeline / gather / transaction so its stages remain recognizable as one unit in `ps` (e.g. `pipeline-<pid>-<n>-stage-0`).

Core integration tests also stopped sharing the fixed names `test`, `t`, and `mybox` across parallel tests: dropped where meaningless, made per-test and pid-suffixed where the hostname assertion needs a known name.

### 2. `test_learn_then_run_collapse` (issue #170)

The test assumed the four `/usr/bin` files it cats share a parent directory after `sandlock learn` canonicalizes observed paths. On the rust-coreutils host, `cat`/`ls`/`env` resolve into `/usr/lib/cargo/bin/coreutils/` while `sh` resolves to dash, so no directory reaches the N=4 collapse threshold, the profile keeps individual file grants, and `run`'s `cat /usr/bin/true` is denied because its canonical target was never observed. Sandlock behaves as designed (the `--merge` hint is the intended UX); the bug is the test's dependence on host layout.

Fix: the test creates its own tempdir with four observed files plus one unobserved file, learns with a pinned `--collapse=4`, and runs `cat` on the unobserved file, which is only readable through the collapsed directory grant.

### 3. `test_collapse_threshold_reads` and `test_collapse_prefix_forces_collapse`

Unit-level twins of the same root cause, made hermetic the same way. The prefix test passed on riscv64 only because a raw exec path happens to remain under `/usr/bin`; it now keeps its file count below the threshold so only `--collapse-prefix` can produce the directory grant.

### 4. `test_config_toml_flag_produces_toml` timeout

Timed out waiting for its sandbox to appear in `ps` while identical sibling spawns passed. The harness discarded the child's stderr and polled blindly with a hard 5s budget, so a startup death and slowness were indistinguishable. Fix: capture stderr, fail fast with exit status and stderr if the child dies, raise the ceiling to 15s. Instrumentation plus margin, not a root-cause fix; a recurrence now produces discriminating evidence.

The SIGPIPE and transaction stderr-tee fixes were split into #173.

Fixes #170

## Test plan

- x86_64: `cargo test --release` for sandlock-core (lib 675 + integration 389, default parallelism), sandlock-cli (82), sandlock-ffi: all pass
- Pre-fix, the core integration suite fails 35/389 on x86_64 with default parallelism, matching the riscv64 sweep, so the pass is attributable to the fix
- Verified the emitted collapse profile grants the parent directory, so the collapse tests cannot pass vacuously
- Needs a confirmation sweep on the riscv64 host (spacemit)

### DIFF
--- a/crates/sandlock-cli/tests/cli_test.rs
+++ b/crates/sandlock-cli/tests/cli_test.rs
@@ -429,21 +429,40 @@ fn sandlock_sleep_args(name: &str) -> Vec<String> {
 
 /// Start a background sandbox, wait for it to be listed by `ps`, then
 /// return its PID. The caller is responsible for killing it.
+///
+/// stderr is captured so wait_for_sandbox can report why a sandbox never
+/// appeared instead of a bare timeout.
 fn spawn_sandbox(name: &str) -> std::process::Child {
     let bin = env!("CARGO_BIN_EXE_sandlock");
     let args = sandlock_sleep_args(name);
     std::process::Command::new(bin)
         .args(&args)
         .stdout(std::process::Stdio::null())
-        .stderr(std::process::Stdio::null())
+        .stderr(std::process::Stdio::piped())
         .spawn()
         .expect("spawn sandlock")
 }
 
+/// Kill `child` if still running and return whatever it wrote to stderr.
+fn drain_child_stderr(child: &mut std::process::Child) -> String {
+    let _ = child.kill();
+    let _ = child.wait();
+    let mut stderr = String::new();
+    if let Some(mut pipe) = child.stderr.take() {
+        use std::io::Read;
+        let _ = pipe.read_to_string(&mut stderr);
+    }
+    stderr
+}
+
 /// Wait for a sandbox to appear in `sandlock ps` output.
-fn wait_for_sandbox(name: &str) -> Result<(), String> {
+///
+/// Fails fast with the child's exit status and stderr if the sandbox process
+/// dies before appearing. The 15s ceiling leaves room for slow hosts running
+/// several sandbox startups in parallel (riscv64 boards).
+fn wait_for_sandbox(child: &mut std::process::Child, name: &str) -> Result<(), String> {
     let bin = env!("CARGO_BIN_EXE_sandlock");
-    for _ in 0..10 {
+    for _ in 0..30 {
         let out = std::process::Command::new(bin)
             .args(["ps"])
             .output()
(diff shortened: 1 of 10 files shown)


congwang-mk:

ghazariann:
> @ghazariann Please take a look. Thanks!

Learn test changes look reasonable to me, private tempdirs over hardcoded `/usr/bin` coreutils is the right fix.
Other hardcoded paths (`/etc/hostname` etc.) are safe: `/etc` files are real files, they don't symlink into a multicall farm.
`dir.path().canonicalize()` is also needed, cause the assertion does exact string matching against `sandlock learn` output, which always canonicalizes paths.

congwang-mk:
Fixed in 1d6e2fe: test_learn_then_run_collapse was the one collapse test still using the raw tempdir path; it now canonicalizes like the other two. It doesn't string-match learn output (it's the learn-then-run behavioral test), so the raw path only worked while /var/tmp resolved to itself, but canonicalizing keeps the learned grant and the run command agreeing on the path regardless.

PR #186, opened 2026-08-04
builder: latch a rejected setter argument instead of coercing it

### LINKED ISSUE #175: An unrecognized BranchAction discriminant silently commits the COW branch
`sandlock_sandbox_builder_on_exit` and `..._on_error` map an unrecognized discriminant to `BranchAction::Commit`, so a binding that passes a bad value gets the change set merged into the workdir instead of an error.

`crates/sandlock-ffi/src/lib.rs:228-232`:

```rust
let action = match action {
    1 => BranchAction::Abort,
    2 => BranchAction::Keep,
    _ => BranchAction::Commit,
};
```

`Commit` is the most destructive of the three: `Abort` and `Keep` both leave the workdir untouched, `Commit` writes to it.

## Reproduction

Verified on `main` (0.8.5, x86-64), driving the C ABI directly so the SDK's enum validation is out of the way — which is exactly the position a hand-written binding is in.

```python
import ctypes, os, shutil, tempfile

lib = ctypes.CDLL("target/release/libsandlock_ffi.so")
P = ctypes.c_void_p

def bfn(name, *extra):
    f = getattr(lib, name); f.restype = P; f.argtypes = [P] + list(extra); return f

b_new     = lib.sandlock_sandbox_builder_new; b_new.restype = P; b_new.argtypes = []
b_read    = bfn("sandlock_sandbox_builder_fs_read", ctypes.c_char_p)
b_workdir = bfn("sandlock_sandbox_builder_workdir", ctypes.c_char_p)
b_storage = bfn("sandlock_sandbox_builder_fs_storage", ctypes.c_char_p)
b_on_exit = bfn("sandlock_sandbox_builder_on_exit", ctypes.c_uint8)

lib.sandlock_sandbox_build.restype = P
lib.sandlock_sandbox_build.argtypes = [P, ctypes.POINTER(ctypes.c_int),
                                       ctypes.POINTER(ctypes.c_char_p)]
lib.sandlock_run.restype = P
lib.sandlock_run.argtypes = [P, ctypes.c_char_p,
                             ctypes.POINTER(ctypes.c_char_p), ctypes.c_uint]

def trial(action_value, label):
    wd, st = tempfile.mkdtemp(), tempfile.mkdtemp()
    b = b_new()
    for p in (b"/usr", b"/lib", b"/lib64", b"/bin"):
        if os.path.exists(p.decode()):
            b = b_read(b, p)
    b = b_workdir(b, wd.encode())
    b = b_storage(b, st.encode())
    b = b_on_exit(b, action_value)

    err, msg = ctypes.c_int(0), ctypes.c_char_p()
    policy = lib.sandlock_sandbox_build(b, ctypes.byref(err), ctypes.byref(msg))
    assert policy, (err.value, msg.value)

    cmd = [b"/bin/sh", b"-c", b"echo written-by-guest > guest.txt"]
    arr = (ctypes.c_char_p * len(cmd))(*cmd)
    lib.sandlock_run(policy, None, arr, len(cmd))

    landed = os.path.exists(os.path.join(wd, "guest.txt"))
    print(f"{label:32} guest changes in workdir: {'YES (committed)' if landed else 'no'}")
    shutil.rmtree(wd, ignore_errors=True); shutil.rmtree(st, ignore_errors=True)

for v, l in [(0, "0 = Commit"), (1, "1 = Abort"), (2, "2 = Keep")]:
    trial(v, l)
for v in (7, 99, 255):
    trial(v, f"{v} = not a valid discriminant")
```

Output:

```
0 = Commit                       guest changes in workdir: YES (committed)
1 = Abort                        guest changes in workdir: no
2 = Keep                         guest changes in workdir: no

7 = not a valid discriminant     guest changes in workdir: YES (committed)
99 = not a valid discriminant    guest changes in workdir: YES (committed)
255 = not a valid discriminant   guest changes in workdir: YES (committed)
```

Every garbage value lands on the one action that modifies the workdir, and `sandlock_sandbox_build` returns success — the caller has no way to notice.

## Why this reads as a missed spot rather than a new class

The same class was already fixed in this file by #92. Two hundred lines below, `crates/sandlock-ffi/src/lib.rs:2152` does the opposite:

```rust
_ => Verdict::Deny,
```

with `policy_ret_unrecognized_values_fail_closed()` at `:2893` pinning it. The `BranchAction` conversion was simply not part of that sweep.

## Why the tests do not catch it

No coverage at all: `on_exit`, `on_error` and `BranchAction` appear nowhere under `crates/sandlock-ffi/tests/`. The core-level branch-action tests construct the enum directly and never cross the ABI, so the conversion is untested from either side.

## Suggested fix

Reject the unknown value instead of choosing for the caller. These setters return the builder and have no error channel, so the cheapest fail-closed option is to map an unrecognized discriminant to `Abort` — the action that cannot destroy anything — and say so in the doc comment. If you would rather have a hard error, the setters need the `err` / `err_msg` treatment `sandlock_sandbox_build` already uses (`crates/sandlock-ffi/src/lib.rs:903`), which is a wider change.

A regression test shaped like `policy_ret_unrecognized_values_fail_closed` would fit next to it.

Happy to send a patch once you pick one.

dzerik:
Closes #175.

Second of three. Stacked on #185, so this branch carries that commit too:
review the second one, `builder: latch a rejected setter argument`, and
merge #185 first.

## The problem

A builder setter returns `Self`, not `Result`, so it has no channel for a
value the core cannot accept. The C ABI answered that by coercing:

- an `on_exit` discriminant with no variant became `Commit`, through the
  fall-through arm of a `match`;
- an unrecognized protection discriminant was a documented no-op;
- every string setter ran its argument through `to_str().unwrap_or("")`,
  except the two mount setters, which dropped the whole call instead.

Each of those runs a configuration the caller never wrote, and says nothing
while doing it. `unwrap_or("")` is reachable without any bug in the caller,
since a path read off `readdir()` is an arbitrary byte string on Linux, and
the empty path it produces is a prefix of every guest path.

## What this does

`SandboxBuilder` carries a pending-error latch. `reject` records a reason a
surface diagnosed itself, `reject_error` records one the core's own parser
produced, and `build()` returns it instead of a `Sandbox`.

The setter contract is otherwise untouched, which is why the bindings do not
move: this commit changes nothing under `go/` or `python/src`.

## Four design decisions worth naming

**The latch holds a `String`, not a `SandboxError`.** `SandboxBuilder` is
`Clone` and `SandboxError` is not; making it `Clone` would widen a public
error type for the benefit of one private field.

**The check sits in `build_unchecked`, not in `build`.** `build_unchecked` is
public and is what `sandlock-oci` calls
(`crates/sandlock-oci/src/policy.rs:463`). A check in `build` alone would let
the one caller that deliberately skips cross-section validation also skip the
caller's own rejected input, which is not the invariant it asked to skip.

**`Clone` carries the latch.** Dropping it there would make `.clone().build()`
a laundering channel for a value the core has already refused.

**First write wins.** The earliest bad input is the one that explains whatever
follows it, so the message names the caller's first mistake rather than its
last.

## What now reports instead of coercing

- `on_exit` and `on_error`, on an unrecognized discriminant. `BranchAction`
  gains `#[repr(u8)]` with explicit discriminants and a `from_repr`, so the
  values the bindings pass as a `u8` are a written-down contract rather than
  the fall-through arm of a `match`. Serde is unaffected: a data-less enum
  serializes by variant name, not by discriminant.

- `allow_degraded` and `disable`, on an unrecognized protection. The no-op was
  documented, which meant a binding built against a newer header was told
  nothing when an older library did not recognize the protection it asked to
  be degradable: the caller believed it had opted out, and the protection
  stayed strict.

- All 22 string setters, through one `setter_arg` helper: a null pointer, and
  bytes that are not UTF-8. Those two stay the C ABI's own verdicts because
  they are representation problems the core cannot see once the value is a
  `&str`; the grammar's verdict still comes from the core untouched. The
  three-argument setters report per half (`env_var key`, `fs_mount_ro host
  path`), so the message names the pointer to fix.

`fs_mount` and `fs_mount_ro` had a coercion of their own shape: a private
`mount_pair` helper answered "add no mount" for a null, non-UTF-8 or empty
path, so the caller who asked for a read-only subtree got a writable one and
the caller who asked for a host directory got nothing there. They go through
the latch now, which is what makes the sentence above true of *every* `*const
c_char` builder setter rather than of most of them.

## Zero and the empty set, in the same commit

The latch stops a surface from inventing a value the caller did not write;
these stop a surface from having to invent a verdict the core would not give.
Both have to be in place before a binding can be reduced to forwarding, and
neither is visible in a binding's own diff.

- **`max_processes = 0`**: the supervisor compares `proc_count >= limit`, so a
  limit of zero denies every fork with `EAGAIN` no matter how few processes
  are alive, and the workload reads "Resource temporarily unavailable" from
  its first subprocess with nothing naming the setting.
- **`num_cpus = 0`**: reaches the synthetic procfs as an empty `/proc/cpuinfo`
  and an affinity mask with no bits, so the guest reads `nproc` = 0.
- **`max_memory = 0`**: zero is the sentinel the supervisor already carries
  for "no ceiling", but the memory handler is registered on `is_some()`. An
  explicit zero therefore installs a ceiling of zero and SIGKILLs the loader's
  first anonymous mmap while `/proc/meminfo` reports the sandbox unlimited.
  Refusing the value is what lets the sentinel keep meaning "unset".
  `max_disk` is deliberately not the same: zero is its documented spelling of
  "unlimited", and one reading is all it has.
- **`cpu_cores = []`**: an affinity mask with no bits, which
  `sched_setaffinity(2)` refuses with `EINVAL`. `confine_child` skipped the
  call for an empty set instead, so the pinning the caller asked for silently
  did not happen and the sandbox ran on every core. Unlike `gpu_devices`,
  where an empty list is the spelling of "every device present", there is no
  cpu set an empty list could stand for.
- **an empty virtual or host path in `fs_mount`/`fs_mount_ro`**: this is the
  check that came back from the C ABI. `mount_pair` was making a policy
  judgement the core's own profile grammar already makes, and making it in the
  one place that could not report it. An empty virtual path is a prefix of
  every guest path, so `ChrootCtx::is_mounted` would match the whole tree and
  short-circuit `can_read` and `can_write`.

## One over-strict check removed

`Confinement::try_from` listed `on_exit` and `on_error` among the fields a
confinement cannot honour. A confinement has no branch to act on: it is
applied in place, and `fs_storage` and `workdir`, the two knobs that create
one, are already refused above it. The check only ever refused a field that
could not have changed the outcome, and it compared against two hardcoded
actions rather than against what `build()` resolves an unset field to, so a
caller who said nothing about the error path was refused a confinement its
policy allowed.

## The C ABI does not move

No signature, no struct and no discriminant value changes.
`include/sandlock.h` changes by 171 lines (141 added, 30 removed) and every
one of them is inside a comment block: with comment lines stripped the header
is byte-identical to its parent. The added ones write the four new refusals
where a binding author reads them, along with `max_open_files = 0`, which was
already refused and had never been documented anywhere.

## Testing

`crates/sandlock-ffi/tests/builder_pending_error.rs` (727 lines) covers the
latch itself: both branch-action setters, survival across later valid calls,
first-write-wins, `Clone`, `build_unchecked`, and null, non-UTF-8 and per-half
arguments across every string setter.

`tests/fs_mount.rs` had four tests pinning the drop-silently behaviour of the
mount setters; they become one that pins the report, over both setters and all
six unusable inputs. In `protection.rs` the two tests that asserted the no-op
now assert the report. `builder.rs` covers `reject_error` directly with a real
`ByteSize::parse` error, asserting the built message is the parser's own text
rather than a doubled `invalid sandbox: invalid sandbox: ...`.

`cargo test -p sandlock-core --lib`: 739 pass at this commit.


congwang-mk:

dzerik:
Agreed, defer it. Answered at length in #185, since the three are one stack and the same answer covers them.

Short version: this PR shows +5745/-975 because it carries #185 underneath it; what is new here is +1570/-367 across 14 files. Still not small, and a release is the wrong moment for it either way. Review order whenever you get to the batch: #185, then this, then #187.

PR #187, opened 2026-08-04
ffi: take sizes and timestamps as strings; both SDKs stop parsing

dzerik:
Third of three. Stacked on #185 and #186, so this branch carries both of
those commits too: review the third one, `ffi: take sizes and timestamps as
strings`, and merge the other two first.

Completes the arc the first two PRs start: after this, neither SDK owns a
grammar. The core owns all of them, and a binding forwards.

## What changes

```
  sandlock_sandbox_builder_max_memory(b, uint64_t bytes)
->                                    (b, const char *size)
  sandlock_sandbox_builder_max_disk(b, uint64_t bytes)
->                                  (b, const char *size)
  sandlock_sandbox_builder_time_start(b, uint64_t epoch_secs)
->                                    (b, const char *timestamp)
```

plus one new symbol, `sandlock_sandbox_builder_time_start_epoch(b, int64_t
seconds, uint32_t nanoseconds)`.

The core parses the strings with the same routines that read
`[limits].memory`, `[limits].disk` and `[determinism].time_start` in a
profile, and `--max-memory`, `--max-disk` and `--time-start` on the command
line. A value it refuses is latched by the mechanism from the previous PR and
comes back from `sandlock_sandbox_build` as the core's own message.

## Why the signatures had to move

"The bindings stop parsing" is not implementable over a `uint64_t` setter: a
binding whose users write `"512M"` has to own a byte-size parser to have
anything to pass. So both of them did. The Go one said what that costs, in its
own words:

```go
// ParseMemory parses a human-friendly size string into bytes. It accepts
// a plain integer (bytes) or a value suffixed with K, M, G, or T (case
// insensitive), e.g. "512M", "1G", "100K". Mirrors the Python SDK's
// parse_memory_size so the two SDKs agree byte-for-byte.
```

Two bindings held in agreement with each other, and neither of them with the
core they were feeding. That parser accepted `"1.5G"`, `"0.5K"`, `"1T"` and
`"512T"`; `ByteSize::parse` takes a decimal integer with a K, M or G suffix
and nothing else, so no flag and no profile has ever accepted any of them. The
entries marked `sdk_only` in `tests/grammar-corpus.json` are exactly that gap.
`ParseTimeStart` had the mirror-image defect on the other grammar: it took a
bare epoch count, which a profile refuses, and refused a pre-1970 stamp, which
a profile takes.

## Why `time_start_epoch` is not a second grammar

It is the numeric door to the same instant, and what needs it is the SDK's own
field: `Sandbox.time_start` takes `str`, `int` or `float`, because an epoch
count is how a caller who starts from `datetime.timestamp()` already holds the
value. A float is not text. With only the string setter, the SDK would have to
render `"1767225600.5"` back into `"2026-01-01T00:00:00.500000000Z"` before it
could forward it: an RFC 3339 *writer* in the binding, one line under the RFC
3339 reader this commit just deleted from it, with its own ways to be wrong
about offsets, leap-second spelling and sub-second digits. The epoch setter
takes the pair the value already is, and `_b_time_start_from` picks the door by
type. Either way the grammar is read exactly once, by the core.

## Core changes that carry it

- `profile::parse_time_start`, `parse_timestamp` and the new
  `time_start_from_epoch` take a label naming the knob that carried the value,
  the way `NetRule::parse_allow` names `--net-allow`, so one grammar can be
  reached from four surfaces and still name the right one in a diagnosis. The
  CLI's private jiff parser is deleted and `sandlock-cli` drops its `jiff`
  dependency.
- `CanonicalTimestamp` gains an `rfc3339` field, for the same reason
  `CanonicalNetRule` carries `spec`. A consumer whose scalar type cannot hold
  both halves was otherwise pushed into a lossy one: a double has about 238ns
  of spacing at 2026 epoch values, so `"...T00:00:00.9999999Z"` rounded up to
  the next whole second and the same profile ran one second later through the
  Python SDK than through the CLI.
- `time.rs`: `calculate_time_offset` resolved both instants with
  `duration_since(UNIX_EPOCH).unwrap_or_default()`, which collapses every
  pre-epoch instant to the epoch. The grammar accepts one
  (`"1969-07-20T20:17:00Z"` parses), so the sandbox ran a clock the caller
  never asked for, 14182980 seconds off, with no error anywhere.

## Bindings

**Go.** `go/internal/policy` is deleted, both files, 131 lines. `MaxMemory`,
`MaxDisk` and `TimeStart` are forwarded verbatim. Every numeric cap is
forwarded whenever it is set, zero included, because the core now has a
verdict on zero for each of them; the fields that cannot express "unset" in
the value itself are therefore pointers, with a `Ptr` helper. `CPUCores` and
`GPUDevices` distinguish nil from empty, so an empty GPU list can still mean
"every GPU present". What remains binding-local is the one verdict Go must
reach on its own: an interior NUL in a Go string, which cannot survive the
conversion to a C string.

**Python.** The size and timestamp fields accept the core's spellings and
forward them; the `__post_init__` guards that used to refuse `"512M"` are
gone. `uid` and `gid` become one `User` value, mirroring the core's `RunAs`:
an unprivileged user namespace maps a single pair, so a uid without a gid is
not a policy the core can apply, and now not a state the SDK can spell.

## BREAKING CHANGES

**Go SDK.** `Name string` -> `*string`; `OnExit`/`OnError` -> `*BranchAction`;
`BranchActionDefault` removed, which renumbers `Commit` 1->0, `Abort` 2->1,
`Keep` 3->2; `MaxProcesses`/`MaxCPU`/`MaxOpenFiles`/`NumCPUs` become pointers;
`UID`/`GID` -> `User *RunAs`; `HTTPPorts []int` -> `[]uint16` (the C ABI
carries a u16, so a wider Go type wrapped silently, 70000 arriving as 4464);
`TimeStart` is RFC 3339 only, in exchange carrying sub-second precision and
pre-1970 instants; `MaxMemory`/`MaxDisk` narrow to the core's grammar; a
non-nil empty `CPUCores` is refused rather than skipped.

**Python SDK.** `uid`/`gid` -> `user: User | None`; `Sandbox.cpu_pct()`
removed; `on_exit`/`on_error` default to `None`; `max_processes` defaults to
`None` (the effective cap is unchanged at 64, it is simply written down once);
`SyscallEvent.category` widens to `str | int`; a missing `/lib64` is no longer
dropped from `fs_readable` behind the caller's back; the size and timestamp
fields accept the core's spellings.

**C ABI.** The three signatures above, plus the additive
`time_start_epoch`. No struct, no discriminant and no other signature moves.

**Canonical profile document.** `CanonicalTimestamp` gains a required
`rfc3339` field, with no serde default on a struct that denies unknown fields,
so an older document no longer deserializes. It loses `Copy` because it now
holds a `String`; `Clone`, `PartialEq` and `Eq` are unaffected.

## The three that are silent at compile time

Most of the list above stops a build or a test run. These do not, and they are
the ones to read before updating:

- **Go `BranchAction` renumbering.** A caller who derives one from a number
  rather than naming a constant, `sandlock.BranchAction(n)` over a stored or
  unmarshalled value, still compiles and now selects a different action, with
  nothing reporting it. Writing the literal straight into the field is caught,
  but only because the same change made that field a pointer, which is an
  accident of packaging rather than a guard.
- **Python `on_error` default.** A policy that says nothing about the error
  path used to abort the branch's writes and now commits them, which is what
  the CLI, a profile and the Go SDK have always done for the same policy. A
  caller relying on the old default gets no warning.
- **`SyscallEvent.category`.** An `event.category == "file"` comparison on an
  unknown category used to be true and is now false.

A C caller passing an integer where a pointer is now expected is a constraint
violation and gets a diagnostic, which is how the signature changes are meant
to be found. The one spelling that slips through is a literal `0`, a valid
null pointer constant: it compiles, and the C ABI then reports it from
`sandlock_sandbox_build` as a NULL argument rather than acting on it.

## One honest regression

Dropping the `/lib64` filter is not in the "fails loudly" group, and calling
it one would be a nicer story than the truth. `build()` does not check that a
readable path exists; nothing in the core does, outside `fs_read_if_exists`,
which nothing on this path calls. The rule is installed in the child, where
Landlock has to open the path, so a policy naming a missing `/lib64` builds,
runs, and comes back as `Result(success=False, error="sandlock_create
failed")` with the path nowhere in it. Under chroot it is skipped without a
word instead.

So this trades a silent drop of one hardcoded path for a loud-but-anonymous
failure on any of them: the right trade for who owns the decision, and a plain
regression in diagnosis. Filter system paths on the way in where the set
varies by host, which is what the MCP default policy and `python/examples` do
now. Making the core name the path it could not grant is worth doing and is
not in this PR.

## Testing

`tests/grammar-corpus.json` is one corpus above all four surfaces, read by
`python/tests/test_setter_grammar_parity.py` (flag, profile, C ABI setter via
ctypes, Python SDK) and `go/grammar_parity_linux_test.go`. A corpus pasted
into each language would be the failure being fixed, one table per surface,
free to drift. `go/core_verdict_linux_test.go` covers the verdicts Go stopped
reaching itself.

`cargo test -p sandlock-core --lib`: 747 pass. Integration: 394. Go: green.
The cbindgen header is regenerated here and matches CI's gate.


congwang-mk:

dzerik:
Agreed, defer it. Answered at length in #185, since the three are one stack and the same answer covers them.

Short version: this PR shows +9874/-1574 because it carries #185 and #186 underneath it; what is new here is +4327/-797 across 46 files, about 70 percent of it tests. This is also the one that carries the breaking changes, so it is the last thing that should go near a release. Review order whenever you get to the batch: #185, then #186, then this.

PR #192, opened 2026-08-07
Issue 183 riscv64 checkpoint restore

sachin2605:
## Summary

Port the checkpoint/restore engine to riscv64, matching x86_64 functionality:
- Restore stub (C): full `#ifdef __riscv64` port with signal frame layout,
  syscall numbers, `sc6()` ecall wrapper, `rt_sigreturn`, and `_start` entry
- Blob builder (Rust): `rearm_restartable_syscall` (a0 return, a7 syscall_nr,
  4-byte ecall rewind) and `build_fpstate_image` (raw passthrough)
- Supervisor (Rust): riscv64 synthetic-image restore test with 32-register
  file, a7+ecall machine code, Sv39-safe addresses
- Build system: detect riscv64 target and try cross-compilers
- Integration test: relax runtime gate so `test_restore_glibc_vdso_program_resumes`
  runs on riscv64
- Unit tests: riscv64 equivalents for all x86_64-gated rearm and fpstate tests

## Files changed

| File | Change |
|------|--------|
| `restore-stub.c` | +157/-30: riscv64 port (signal frame, mmap, dup3, rt_sigreturn, _start) |
| `resume.rs` | +136/-8: generalize cfg gate, add riscv64 synthetic-image test |
| `restore_blob.rs` | +91/-17: rearm_restartable_syscall, build_fpstate_image, +3 riscv64 tests |
| `build.rs` | +23/-1: riscv64 cross-compiler detection |
| `sandbox.rs` | +4/-1: relax restore gate |
| `test_restore.rs` | +4/-2: relax runtime gate |

## Key design decisions

- **Blob format is arch-agnostic**: same `BLOB_MAGIC`, `VERSION`, `HEADER_LEN`;
  register count determined by `regs_len`; both arches are LE
- **ecall is 4 bytes → PC rewind by 4** (vs x86_64's 2-byte syscall)
- **dup3** (syscall 24) replaces **dup2** on riscv64
- **No xstate framing**: kernel stores `__riscv_d_ext_state` directly in
  `sc_fpregs` — no separate `fp_buf`, no magic framing
- **Signal frame**: `uc_mcontext` at offset 0xA8=168; `sc_regs[32]` is 1:1
  with `user_regs_struct` — no remapping, no `CSGSFS`, no `arch_prctl`
- **Sv39 user space ends at 256 GiB** — synthetic test addresses kept below
  `0x40_0000_0000`

### DIFF
--- a/crates/sandlock-core/build.rs
+++ b/crates/sandlock-core/build.rs
@@ -34,20 +34,35 @@ fn main() {
     let stub_src = manifest_dir.join("src/checkpoint/restore-stub.c");
     let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
     let stub_bin = out_dir.join("restore-stub");
+    let host = std::env::var("HOST").unwrap_or_default();
     let target = std::env::var("TARGET").unwrap_or_default();
-    let (ccs, warn) = if target.starts_with("riscv64")
-        || target.starts_with("riscv64gc")
-    {
-        (
-            &["riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..],
-            "cannot compile restore-stub for riscv64: its restore tests will be \
-             skipped. Install a riscv64 cross-compiler (e.g. riscv64-linux-gnu-gcc).",
-        )
+    let is_riscv64 = target.starts_with("riscv64");
+    let (ccs, warn) = if is_riscv64 {
+        if host.starts_with("riscv64") {
+            (
+                &["cc", "riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..],
+                "cannot compile restore-stub for riscv64: its restore tests will be \
+                 skipped. Install gcc.",
+            )
+        } else {
+            (
+                &["riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..],
+                "cannot compile restore-stub for riscv64: its restore tests will be \
+                 skipped. Install a riscv64 cross-compiler (e.g. riscv64-linux-gnu-gcc).",
+            )
+        }
     } else {
         (
             &["cc"][..],
             "cannot compile restore-stub: its restore tests will be skipped.",
         )
+    };
+    // The link address must match restore_blob::STUB_BASE and must sit below
+    // the Sv39 user ceiling (256 GiB) on riscv64.  x86_64 uses 0x300_0000_0000.
+    let text_segment = if is_riscv64 {
+        "-Wl,-Ttext-segment=0x3000000000"
+    } else {
+        "-Wl,-Ttext-segment=0x30000000000"
     };
     build_static(
         &stub_src,
@@ -58,19 +73,12 @@ fn main() {
             "-nostdlib",
             "-no-pie",
             "-O2",
-            // Without these, loop-idiom recognition rewrites the stub's own
-            // hand-written memset/memcpy bodies into calls to memset/memcpy,
-            // i.e. into infinite self-recursion. There is no libc to fall back
-            // on, so the stub must keep its byte loops as byte loops.
             "-ffreestanding",
             "-fno-tree-loop-distribute-patterns",
-            "-Wl,-Ttext-segment=0x30000000000",
+            text_segment,
         ],
         warn,
     );
-    // Emit the path every run (rustc-env is not cached across build-script runs),
-    // whether or not the binary was just (re)built.
-    println!("cargo:rustc-env=RESTORE_STUB_PATH={}", stub_bin.display());
 }
 
 /// Compile `src` to `bin` with the first working compiler in `ccs`, skipping the
--- a/crates/sandlock-core/build.rs
+++ b/crates/sandlock-core/build.rs
@@ -27,13 +27,32 @@ fn main() {
     // text and stack have to sit outside the address range programs occupy. The
     // default -no-pie base (0x400000) is exactly where a static ET_EXEC workload
     // loads, so the checkpoint's own text would be mapped over the running stub.
+    //
+    // Cross-compilation: when TARGET is riscv64gc-unknown-linux-gnu (or any
+    // riscv64* variant), look for a riscv64 cross-compiler. On the host it
+    // uses plain `cc` as before.
     let stub_src = manifest_dir.join("src/checkpoint/restore-stub.c");
     let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
     let stub_bin = out_dir.join("restore-stub");
+    let target = std::env::var("TARGET").unwrap_or_default();
+    let (ccs, warn) = if target.starts_with("riscv64")
+        || target.starts_with("riscv64gc")
+    {
+        (
+            &["riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..],
+            "cannot compile restore-stub for riscv64: its restore tests will be \
+             skipped. Install a riscv64 cross-compiler (e.g. riscv64-linux-gnu-gcc).",
+        )
+    } else {
+        (
+            &["cc"][..],
+            "cannot compile restore-stub: its restore tests will be skipped.",
+        )
+    };
     build_static(
         &stub_src,
         &stub_bin,
-        &["cc"],
+        ccs,
         &[
             "-static",
             "-nostdlib",
@@ -47,7 +66,7 @@ fn main() {
             "-fno-tree-loop-distribute-patterns",
             "-Wl,-Ttext-segment=0x30000000000",
         ],
-        "cannot compile restore-stub: its restore tests will be skipped.",
+        warn,
     );
     // Emit the path every run (rustc-env is not cached across build-script runs),
     // whether or not the binary was just (re)built.
--- a/crates/sandlock-core/build.rs
+++ b/crates/sandlock-core/build.rs
@@ -79,6 +79,9 @@ fn main() {
         ],
         warn,
     );
+    // Emit the path every run (rustc-env is not cached across build-script runs),
+    // whether or not the binary was just (re)built.
+    println!("cargo:rustc-env=RESTORE_STUB_PATH={}", stub_bin.display());
 }
 
 /// Compile `src` to `bin` with the first working compiler in `ccs`, skipping the
--- a/crates/sandlock-core/build.rs
+++ b/crates/sandlock-core/build.rs
@@ -8,14 +8,17 @@ fn main() {
 
     // rootfs-helper: an ordinary static-libc test fixture (chroot tests). It
     // lives in tests/ and its binary sits beside it (a git-ignored artifact).
-    build_static(
+    if !build_static(
         &repo_root.join("tests/rootfs-helper.c"),
         &repo_root.join("tests/rootfs-helper"),
         &["musl-gcc", "cc"],
         &["-static", "-O2"],
-        "cannot compile tests/rootfs-helper: chroot tests will fail. \
-         Install musl-tools or static libc.",
-    );
+    ) {
+        println!(
+            "cargo:warning=cannot compile tests/rootfs-helper: chroot tests will \
+             fail. Install musl-tools or static libc."
+        );
+    }
 
     // restore-stub: a core component of the restore engine (the supervisor execs
     // it to reconstruct a checkpoint), freestanding, no libc, no PIE. It lives
@@ -37,24 +40,29 @@ fn main() {
     let host = std::env::var("HOST").unwrap_or_default();
     let target = std::env::var("TARGET").unwrap_or_default();
     let is_riscv64 = target.starts_with("riscv64");
-    let (ccs, warn) = if is_riscv64 {
+    // Checkpoint restore is claimed only on x86_64 and riscv64 (see
+    // `restore_interactive`); on those arches a stub build failure is fatal, not
+    // a silent skip — a green build with no stub is how regressions slip past CI.
+    let is_restore_arch = target.starts_with("x86_64") || is_riscv64;
+    let (ccs, fail_msg) = if is_riscv64 {
         if host.starts_with("riscv64") {
             (
                 &["cc", "riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..],
-                "cannot compile restore-stub for riscv64: its restore tests will be \
-                 skipped. Install gcc.",
+                "failed to compile restore-stub for riscv64: no working C compiler \
+                 (install gcc); checkpoint restore is unavailable",
             )
         } else {
             (
                 &["riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..],
-                "cannot compile restore-stub for riscv64: its restore tests will be \
-                 skipped. Install a riscv64 cross-compiler (e.g. riscv64-linux-gnu-gcc).",
+                "failed to compile restore-stub for riscv64: no working cross-compiler \
+                 (install riscv64-linux-gnu-gcc); checkpoint restore is unavailable",
             )
         }
     } else {
         (
             &["cc"][..],
-            "cannot compile restore-stub: its restore tests will be skipped.",
+            "failed to compile restore-stub: no working C compiler \
+             (install cc/gcc); checkpoint restore is unavailable",
         )
     };
     // The link address must match restore_blob::STUB_BASE and must sit below
@@ -64,7 +72,7 @@ fn main() {
     } else {
         "-Wl,-Ttext-segment=0x30000000000"
     };
-    build_static(
+    if !build_static(
         &stub_src,
         &stub_bin,
         ccs,
@@ -77,26 +85,32 @@ fn main() {
             "-fno-tree-loop-distribute-patterns",
             text_segment,
         ],
-        warn,
-    );
+    ) {
+        if is_restore_arch {
+            panic!("{fail_msg}");
+        }
+        println!("cargo:warning={fail_msg}");
+    }
     // Emit the path every run (rustc-env is not cached across build-script runs),
     // whether or not the binary was just (re)built.
     println!("cargo:rustc-env=RESTORE_STUB_PATH={}", stub_bin.display());
 }
 
 /// Compile `src` to `bin` with the first working compiler in `ccs`, skipping the
-/// work when `bin` is newer than `src`. Emits `warn` (as a cargo warning) if no
-/// compiler succeeds. A missing source is silently skipped (packaged crate).
-fn build_static(src: &Path, bin: &Path, ccs: &[&str], args: &[&str], warn: &str) {
+/// work when `bin` is newer than `src`. Returns `false` only when the source is
+/// present, newer than `bin`, and no compiler in `ccs` succeeded; a missing
+/// source (a packaged crate) or an up-to-date `bin` reports success. The caller
+/// decides whether that failure is a hard error or a warning.
+fn build_static(src: &Path, bin: &Path, ccs: &[&str], args: &[&str]) -> bool {
     println!("cargo:rerun-if-changed={}", src.display());
     if !src.exists() {
-        return;
+        return true;
     }
     if bin.exists() {
         if let (Ok(s), Ok(b)) = (src.metadata(), bin.metadata()) {
             if let (Ok(st), Ok(bt)) = (s.modified(), b.modified()) {
                 if bt >= st {
-                    return;
+                    return true;
                 }
             }
         }
@@ -111,8 +125,8 @@ fn build_static(src: &Path, bin: &Path, ccs: &[&str], args: &[&str], warn: &str)
             .map(|s| s.success())
             .unwrap_or(false);
         if ok {
-            return;
+            return true;
         }
     }
-    println!("cargo:warning={warn}");
+    false
 }
--- a/crates/sandlock-core/src/checkpoint/restore-stub.c
+++ b/crates/sandlock-core/src/checkpoint/restore-stub.c
@@ -1,5 +1,5 @@
 /*
- * restore-stub: freestanding self-restore stub (x86_64).
+ * restore-stub: freestanding self-restore stub (x86_64, riscv64).
  *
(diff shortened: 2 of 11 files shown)


crates/sandlock-core/src/checkpoint/restore-stub.c
```diff
@@ -193,11 +240,47 @@ enum { UR_R15=0,UR_R14,UR_R13,UR_R12,UR_RBP,UR_RBX,UR_R11,UR_R10,UR_R9,UR_R8,
...
+/* riscv64: rt_sigreturn reads frame at sp = (struct rt_sigframe *)sp.
+ * siginfo (128 bytes) + ucontext. uc_mcontext is at uc+0xA8 (168).
+ * sc_regs[32] at sigcontext+0x00, sc_fpregs at sigcontext+0x100.
+ * The gp[] blob order is ptrace: pc=0, ra=1, sp=2, gp=3, tp=4, t0-t2=5-7,
+ * s0-s1=8-9, a0-a7=10-17, s2-s11=18-27, t3-t6=28-31 — a 1:1 mapping to
+ * sc_regs[32], so no remap is needed. tp (thread pointer) is carried by the
+ * signal frame's sc_regs[4]; nothing needs arch_prctl. */
+struct sigctx {
+    u64 gregs[32];           /* sc_regs: 32 gregs, 256 bytes */
+    u8  fpregs[544];         /* sc_fpregs: union __riscv_fp_state (max Q ext) */
```
congwang-mk: **Blocker: this branch does not compile — `u8` is never declared.**

The typedefs at lines 136-138 are `u64`, `u32`, `i64` only; `u8` was never needed before this PR. Building the stub natively on riscv64 (Debian gcc 14, `mmu: sv39`) with exactly the flags `build.rs` passes:

```
/tmp/stub.c:254:5: error: unknown type name 'u8'
  254 |     u8  fpregs[544];
/tmp/stub.c:265:5: error: unknown type name 'u8'
/tmp/stub.c:271:5: error: unknown type name 'u8'
```

Adding `typedef unsigned char u8;` next to the others makes it compile and link.

This is why CI is green while none of the riscv64 work has ever run: `build_static` turns a compiler failure into a `cargo:warning`, and every riscv64 test then early-returns on `!stub.exists()`.

crates/sandlock-core/src/checkpoint/restore-stub.c
```diff
@@ -193,11 +240,47 @@ enum { UR_R15=0,UR_R14,UR_R13,UR_R12,UR_RBP,UR_RBX,UR_R11,UR_R10,UR_R9,UR_R8,
...
+struct uctx {
+    u64 uc_flags;            /* 0x00 */
+    u64 uc_link;             /* 0x08 */
+    u64 ss_sp;               /* 0x10 */
+    u32 ss_flags;            /* 0x18 */
+    u32 _pad;                /* 0x1C */
+    u64 ss_size;             /* 0x20 */
+    u64 uc_sigmask;          /* 0x28 */
+    u8  __unused[120];       /* 0x30 */
+    struct sigctx mc;        /* 0xA8 (168) */
```
congwang-mk: **Blocker: `uc_mcontext` is at offset 176 (0xB0), not 168 (0xA8).**

`struct sigcontext` has 16-byte alignment on riscv64 — both `__riscv_q_ext_state.f[64]` and `__riscv_extra_ext_header.__padding` carry `__attribute__((aligned(16)))`, so the union and therefore `struct sigcontext` are 16-aligned. 168 is not a multiple of 16, so the kernel pads to 176. Measured on a riscv64 host against `<asm/ucontext.h>`:

```
offsetof(kernel struct ucontext, uc_mcontext) = 176 (0xb0)
sizeof(struct sigcontext) = 784
```

`struct sigctx` here has alignment 8, so the compiler places `mc` at exactly 168 and the whole register file lands 8 bytes low. `rt_sigreturn` then reads `pc` from `gregs[1]` (`ra`), `sp` from `gregs[2]` (`gp`), and so on. In the new synthetic test `ra` is 0, so the restored process jumps to 0 and dies.

Fix: `__attribute__((aligned(16)))` on `struct sigctx` (or widen `__unused` to 128 bytes and keep the comment honest).

crates/sandlock-core/src/checkpoint/restore-stub.c
```diff
@@ -61,6 +62,21 @@
 #define STUB_BASE 0x30000000000UL
 #define STUB_SPAN 0x400000UL
 
+#ifdef __riscv
```
congwang-mk: `__riscv` is also defined on RV32, where `__riscv_xlen == 32`. The `u64`-based `struct sigctx`/`struct uctx` mirrors, the `SYS_*` numbers, and the 32x8-byte `sc_regs` copy are all RV64-only, and the `#else #error` added below never fires there — so an RV32 build would compile silently wrong rather than being rejected.

Suggest `#if defined(__riscv) && __riscv_xlen == 64`.

crates/sandlock-core/src/checkpoint/restore-stub.c
```diff
@@ -414,13 +505,41 @@ static void _start_c(u64 *sp) {
...
+     * The FP state is embedded inline in sc_fpregs (no pointer indirection,
+     * no magic framing — restore_blob.rs sends the raw __riscv_d_ext_state). */
+    struct rt_sf sf;
+    memset(&sf, 0, sizeof sf);
+    /* gp has regs_len / 8 entries; copy all of them into sc_regs[32]. */
+    u32 nregs = h->regs_len / 8;
+    if (nregs > 32) nregs = 32;
+    memcpy(sf.uc.mc.gregs, gp, nregs * sizeof(u64));
+    if (h->fpstate_len) {
+        if (h->fpstate_len > sizeof(sf.uc.mc.fpregs))
```
congwang-mk: Two things here:

1. This clamp is dead. Line 349 already does `if (h->fpstate_len > FP_MAX) die(3);`, and `FP_MAX` is 544 on riscv64 — the same bound. It also mutates the header in place inside `ctrl_buf`, which the rest of the stub treats as read-only after validation.

2. If it were reachable, 544 is the wrong bound. The kernel's second `sigcontext` member is `union { union __riscv_fp_state sc_fpregs; struct __riscv_extra_ext_header sc_extdesc; }`, which is 528 bytes, with `sc_extdesc.reserved` at offset 516 and `sc_extdesc.hdr` at 520 (verified: `sizeof(struct sigcontext) == 784` = 256 + 528). A 544-byte copy would overwrite `reserved`, and `restore_sigcontext()` returns `-EINVAL` on a non-zero `reserved`, so `rt_sigreturn` would fail instead of resuming.

Not reachable today (riscv `NT_PRFPREG` yields 264 bytes), but `FP_MAX` should be 516/528 rather than 544, and the comment above it calls 544 the Q-extension size, which it isn't.

crates/sandlock-core/src/checkpoint/restore_blob.rs
```diff
@@ -374,6 +374,21 @@ fn rearm_restartable_syscall(regs: &mut [u64]) {
...
+#[cfg(target_arch = "riscv64")]
+fn rearm_restartable_syscall(regs: &mut [u64]) {
+    // riscv64 user_regs_struct layout indices (32 u64):
+    //   pc, ra, sp, gp, tp, t0-t2, s0-s1, a0-a7, s2-s11, t3-t6
+    const A0: usize = 10;
+    const A7: usize = 17;
+    const PC: usize = 0;
+    if let (Some(&a0), Some(&a7)) = (regs.get(A0), regs.get(A7)) {
+        if matches!(a0 as i64, -512 | -513 | -514 | -516) {
+            regs[A0] = a7;                         // restore original syscall number
```
congwang-mk: **Blocker: on riscv64 `a0` is the first *argument*, not the syscall number, so this corrupts the restarted call.**

The x86_64 version works because `rax` carries both the number and the return value. riscv64 splits them: the number stays in `a7` (the kernel never clobbers it) and `a0` is overwritten by the return value. The kernel's own fixup is `arch/riscv/kernel/signal.c`:

```c
case -ERESTARTNOHAND:
case -ERESTARTSYS:
case -ERESTARTNOINTR:
case -ERESTART_RESTARTBLOCK:
        regs->a0 = regs->orig_a0;
        regs->epc = restart_addr;
        break;
```

Concretely: a checkpoint taken while blocked in `read(5, buf, n)` has `a0 = -ERESTARTSYS`, `a7 = 63`. After this re-arm the `ecall` re-executes as `read(63, buf, n)` — EBADF, or silently the wrong fd.

The harder problem is that `orig_a0` is not recoverable from `ps.regs`: `riscv_gpr_get()` writes only `sizeof(struct user_regs_struct)` (256 bytes, `pc`..`t6`), and `orig_a0` sits past `status`/`badaddr`/`cause` in `pt_regs`. So the re-arm as designed cannot be implemented from the captured register file — it needs an extra capture (e.g. `/proc/<pid>/syscall`, which does expose the original arg0) or should be left unimplemented on riscv64 rather than writing a wrong `a0`.

Two related notes:
- For `-ERESTART_RESTARTBLOCK` the kernel sets `a7 = __NR_restart_syscall`; it does not re-run the original number.
- The new unit test at line 911 asserts `regs[10] == 34`, i.e. it locks in this behaviour, so it will need to change with the fix.

crates/sandlock-core/build.rs
```diff
@@ -27,13 +27,32 @@ fn main() {
...
+    // uses plain `cc` as before.
     let stub_src = manifest_dir.join("src/checkpoint/restore-stub.c");
     let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
     let stub_bin = out_dir.join("restore-stub");
+    let target = std::env::var("TARGET").unwrap_or_default();
+    let (ccs, warn) = if target.starts_with("riscv64")
+        || target.starts_with("riscv64gc")
+    {
+        (
+            &["riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..],
```
congwang-mk: **Blocker: a *native* riscv64 build loses the compiler entirely.**

`TARGET` is `riscv64gc-unknown-linux-gnu` for a plain `cargo build` on a riscv64 host too, so the candidate list becomes just these two cross-prefixed names — neither of which exists on a riscv64 machine, where the compiler is `/usr/bin/cc` / `/usr/bin/gcc` (checked on our riscv64 test host). The stub is then never built, `stub_links_at_the_reserved_base`, the new synthetic test, and `test_restore_glibc_vdso_program_resumes` all silently return early, and `restore_interactive` fails at runtime with "restore-stub was not built" — on exactly the platform this PR is for.

Keep `cc` in the list, gated on `HOST` so a genuine cross build doesn't silently produce a host binary:

```rust
let host = std::env::var("HOST").unwrap_or_default();
// ... riscv64 branch:
if host.starts_with("riscv64") {
    &["cc", "riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..]
} else {
    &["riscv64-linux-gnu-gcc", "riscv64-unknown-linux-gnu-gcc"][..]
}
```

Minor, same hunk: `|| target.starts_with("riscv64gc")` on line 39 is subsumed by `target.starts_with("riscv64")`.

crates/sandlock-core/build.rs
```diff
@@ -47,7 +66,7 @@ fn main() {
             "-fno-tree-loop-distribute-patterns",
             "-Wl,-Ttext-segment=0x30000000000",
```
congwang-mk: **Blocker: 3 TiB is above the Sv39 user-address ceiling, so the stub cannot be exec'd on real riscv64 hardware.**

`restore_blob::STUB_BASE` is `0x300_0000_0000` and this link address matches it, but Sv39 (what JH7110, SpacemiT K1, and most riscv64 Linux boards run) gives `TASK_SIZE` = 256 GiB. The PR's own new comment in `resume.rs` says addresses must stay below `0x40_0000_0000`; this is 12x over that.

Verified on a riscv64 host (`mmu: sv39`, `[stack]` at `0x3ff1469000`):

```
mmap STUB_BASE 0x300_0000_0000  -> ENOMEM
mmap test CODE 0x200_0000_0000  -> ENOMEM
mmap            0x2000_0000     -> ok

$ ./restore-stub          # ET_EXEC with p_vaddr = 0x300_0000_0000
Segmentation fault        # elf_map() fails past the point of no return
```

Relinking the *same* source at `-Wl,-Ttext-segment=0x3000000000` (192 GiB, under the ceiling) makes it exec and run correctly — it reaches the blob parser and exits 3 on a short blob, as designed.

So riscv64 needs its own stub base inside the Sv39 window, chosen here and in `restore_blob.rs:45` together (the `stub_links_at_the_reserved_base` test exists precisely to catch them drifting apart, and it will once the stub actually builds).

Aside, since it's the obvious suspicion with a 3 TiB text segment: `-mcmodel` is *not* a problem. GCC on riscv64 Linux defaults to `medany`, and `la sp, stub_stack` assembles to `auipc`/`ld` through the GOT — the link succeeds at either base.

crates/sandlock-core/src/checkpoint/resume.rs
```diff
@@ -572,4 +572,138 @@ mod tests {
...
+        //   ecall
+        //   addi a7, zero, 93        # __NR_exit  (a0 still holds write's ret)
+        //   ecall
+        let mut code_page = vec![0u8; PAGE as usize];
+        {
+            let c = &mut code_page;
+            let mut w = 0usize;
+            let mut put = |bytes: &[u8]| { c[w..w + bytes.len()].copy_from_slice(bytes); w += bytes.len(); };
+            put(&0x00A00513u32.to_le_bytes()); // addi a0, zero, 10
+            put(&0x200005B7u32.to_le_bytes()); // lui  a1, 0x20000
```
congwang-mk, line 616: **Blocker: this pair does not materialize `CODE + 64`.**

`lui a1, 0x20000` puts `0x20000` in bits 31:12, giving `a1 = 0x2000_0000`; the `addi` makes it `0x2000_0040`. But `CODE + 64` is `0x200_0000_0040` — `lui`+`addi` reach 32 bits, and `CODE` is 2^41. The restored payload calls `write(10, 0x2000_0040, 1)` against an unmapped address, gets `-EFAULT`, and the test fails at `assert_eq!(n, 1, ...)`.

The encoded instructions are self-consistent with `CODE = 0x2000_0000`, so it looks like the constant on line 588 picked up three extra zeros.

That also fixes a second problem in the same test: `0x200_0000_0000` is 2 TiB, which is 8x above the Sv39 ceiling the comment on line 585 cites, and `mmap` there returns ENOMEM on Sv39 hardware (measured). `const CODE = 0x2000_0000` / `const STACK = 0x2001_0000` satisfies both the encoding and the comment.

congwang-mk (commented):
Reviewed the riscv64 port. The direction is right and the ABI research (1:1 `sc_regs`/ptrace ordering, no `arch_prctl`, no xstate framing, dup3) is accurate, but the code has not run yet: **the riscv64 stub does not compile**, and CI cannot tell you that because `build_static` downgrades a compiler failure to a `cargo:warning` and every riscv64 test then skips on `!stub.exists()`. The `Rust cross-build (riscv64)` job is green for the same reason.

I checked the findings below on a real riscv64 host (Sv39, 6.18.3) and against the kernel uapi headers rather than by reading alone.

Six blockers, in the order you'd hit them:

1. `restore-stub.c` — `u8` is never typedef'd; the riscv64 branch fails to compile.
2. `restore-stub.c` — `uc_mcontext` is at 176, not 168; `rt_sigreturn` restores a register file shifted by 8 bytes.
3. `build.rs` / `restore_blob.rs:45` — the 3 TiB `STUB_BASE` is above the Sv39 user ceiling; the stub SIGSEGVs at exec on real boards.
4. `resume.rs` — the synthetic payload's `lui`/`addi` yields `0x2000_0040`, not `CODE + 64`.
5. `restore_blob.rs` — the restart re-arm writes the syscall number into `a0`, which is the first *argument* on riscv64; and `orig_a0` isn't reachable through `NT_PRSTATUS`.
6. `build.rs` — a native riscv64 build drops `cc` from the compiler list, so the stub is never built there at all.

Plus smaller things inline: the `#ifdef __riscv` guard also matches RV32, and the 544-byte `sc_fpregs` bound overshoots the kernel's 528-byte union (the clamp itself is dead code).

One doc nit outside the diff: `sandbox.rs:1067` still says "x86_64 restore engine only." in the `restore_interactive` doc comment, three lines above the gate this PR widened.

Since none of this can be caught by the current CI shape, it's worth making the stub build failure loud on the arches we claim to support — a `cargo:warning` is invisible in a green build, which is how all six of these got this far.

congwang-mk:
Blocker: riscv64 restore is now publicly enabled, but restartable syscalls are not handled. restore_blob.rs:388 explains the limitation, and restore_blob.rs:483 makes rearm_restartable_syscall a no-op for riscv64. That means if checkpoint capture interrupts nanosleep, futex, read, etc., the restored process resumes after ecall with a0 = -ERESTART* instead of a valid syscall result or a rearmed syscall. This is exactly the class x86_64 fixes at restore_blob.rs:374. Since sandbox.rs:1082 now allows riscv64 restore, this can produce corrupted resumes for ordinary workloads. Either keep riscv64 restore gated off, reject riscv64 checkpoints whose a0 contains a restart sentinel, or implement a real riscv64 restart strategy.

congwang-mk:
@sachin2605 Any comment/update?

sachin2605:
@congwang-mk  please suggest in the following. 

Two things make a real riscv64 rearm hard.

 - On riscv64 a0 is both the first argument and the return value; the sentinel has already overwritten the original
   first argument, and orig_a0 is not exposed by PTRACE_GETREGSET/NT_PRSTATUS (unlike x86_64's orig_rax, which is why
   x86_64 can re-arm). Confirmed in the kernel uapi struct user_regs_struct — 32 regs, no orig_a0.
 - ERESTART_RESTARTBLOCK also needs restart_block (remaining timeout), which is kernel state and isn't captured.

 Key finding that changes the recommendation

 The behavior is kernel-version dependent, and I verified this in the kernel source:

 - Commit ce4f78f1b53d ("riscv: signal: handle syscall restart before get_signal", Linux ~6.6) moved riscv64 to
   arm64-style: arch_do_signal_or_restart now runs the restart fixup before get_signal/ptrace_stop — the code comment
   is literally "so that a debugger will see the already changed PC".
 - Before 6.6, riscv64 re-armed after the ptrace stop (like x86_64) — so the sentinel is in the captured a0, and the
   reviewer's corruption is real.
 - Since 6.6, the re-arm has already happened at capture: a0 = orig_a0, pc rewound by 4. The captured register file is
   already correct, and no restore-time rearm is needed.

 So a correct "real restart" strategy would only be needed for pre-6.6 kernels, where it's blocked by the missing
 orig_a0 anyway. That makes the "real restart" option low-value and high-risk.

  ### Recommendation: reject checkpoints whose a0 holds a restart sentinel (reviewer's option 2)

 This is the robust, version-agnostic fix and doesn't regress the feature:

 - On ≥6.6 kernels the check essentially never fires (kernel pre-rearms), so riscv64 restore stays fully usable.
 - On <6.6 kernels it refuses instead of corrupting, with a clear error.

 Concretely, at the ptrace stop, if a0 (regs[10], per restore-stub.c:255-256 order pc=0…a0-a7=10-17) is one of
 -512|-513|-514|-516 (never -515, which is ENOIOCTLCMD), reject.

 ### Why not the other options

 - Gate riscv64 restore off — regresses the entire point of the PR.
 - Real riscv64 restart strategy — needs orig_a0, which ptrace doesn't expose; would require reading
   /proc/<pid>/syscall at capture (fragile, and still can't do ERESTART_RESTARTBLOCK correctly) or a
   seccomp/syscall-entry hook. High risk, and moot on ≥6.6 kernels where the kernel already re-arms.

sachin2605:
updated following 
Review body + 8 inline comments (@congwang-mk, 2026-08-07) — six blockers, smaller items, one doc nit, one
 suggestion:

 +---+------------------------------------------+--------------------------------------------------------------------+
 | # | Comment                                  | Status                                                             |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | 1 | u8 never typedef'd → stub won't compile  | ✅ typedef unsigned char u8; (restore-stub.c:144)                  |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | 2 | uc_mcontext at 176, not 168              | ✅ __unused[128], mc at 0xB0 (restore-stub.c:272-273)              |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | 3 | 3 TiB STUB_BASE above Sv39 ceiling       | ✅ riscv64 0x30_0000_0000 (192 GiB) in restore_blob.rs, build.rs,  |
 |   |                                          | restore-stub.c                                                     |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | 4 | lui/addi → 0x2000_0040, not CODE+64      | ✅ CODE = 0x2000_0000 (resume.rs:588)                              |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | 5 | rearm writes syscall number into a0      | ✅ reject instead (my 883f946)                                     |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | 6 | native riscv64 build drops cc            | ✅ host.starts_with("riscv64") keeps cc (build.rs:48)              |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | — | __riscv matches RV32                     | ✅ && __riscv_xlen == 64 everywhere                                |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | — | FP_MAX 544 overshoots 528-byte union;    | ✅ FP_MAX 516, fpregs[528], clamp removed                          |
 |   | dead clamp                               |                                                                    |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | — | sandbox.rs "x86_64 only" doc             | ✅ updated                                                         |
 +---+------------------------------------------+--------------------------------------------------------------------+
 | — | make stub build failure loud             | ✅ fixed this turn (build.rs)                                      |
 +---+------------------------------------------+--------------------------------------------------------------------+

 Issue comments:
 - Restartable-syscall blocker (2026-08-10) → ✅ addressed by 883f946.
 - "Any comment/update?" (2026-08-22) → ✅ answered by the reject fix + the existing reply.

PR #203, opened 2026-08-23
checkpoint: riscv64 post-merge review follow-ups

congwang-mk:
Two follow-ups from the post-merge review of #192.

**Assert FP_MAX fits sc_fpregs in the restore stub.** The riscv64 sc_fpregs memcpy has no bound of its own: it is safe because validate rejects fpstate_len > FP_MAX and FP_MAX (516) fits the 528-byte field, but nothing enforced that relation. Raising FP_MAX or shrinking the struct would silently turn the copy into a stack overflow inside the signal frame, so a _Static_assert now pins it at the struct definition.

**Share the riscv64 restart-sentinel rejection.** Capture and rearm_restartable_syscall each carried their own copy of the sentinel check plus a near-identical multi-line error, and the two messages had already drifted. Both now route through one reject_restart_sentinel helper so the check and its wording cannot diverge again.

Verified locally: the stub compiles for riscv64 (clang -fsyntax-only cross-check, which also proves the assert holds) and the 38 checkpoint lib tests pass on x86_64. The riscv64 Rust paths rely on the cross-build CI job, as usual on this machine.

### DIFF
--- a/crates/sandlock-core/src/checkpoint/capture.rs
+++ b/crates/sandlock-core/src/checkpoint/capture.rs
@@ -423,14 +423,8 @@ pub(crate) fn capture(pid: i32, policy: &Sandbox) -> Result<Checkpoint, Sandlock
         // while the tell-tale sentinel is still in `a0`. This saves the memory
         // dump and surfaces the error at the actionable moment.
         #[cfg(target_arch = "riscv64")]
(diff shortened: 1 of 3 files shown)


congwang-mk:
@sachin2605 Please take a look. Thanks

sachin2605:
@congwang-mk  Looks good to me. Thanks for adding 👍

PR #207, opened 2026-09-02
control: replace the /dev/shm registry with abstract Unix sockets

congwang-mk:
## Summary

Replaces the `/dev/shm/sandlock-$UID/<name>/` runtime directory (pid file, mode marker, named socket) with two abstract Unix sockets per sandbox, `\0sandlock/<uid>/<name>` and `\0sandlock/<uid>/<name>/pgrp`, both bound by the supervisor before it forks. The first commit drops the `control_socket` opt-out knob: whether a sandbox is visible to `sandlock ps` is the operator's concern, not a policy field, and the knob silently skipped the name uniqueness check.

Why:

- **Nesting needs no outer grant.** A nested sandlock only needs permission to create an AF_UNIX socket, not a writable /dev/shm. The Landlock abstract-socket scope restricts connecting *out* of a domain; a host-side `sandlock ps` still reaches an inner supervisor.
- **Nothing is ever stale.** The names vanish with the supervisor, so `ps` no longer prunes directories, probes pids, or guards against setup races.
- **Uniqueness is enforced by the kernel.** `bind` on a taken name fails with EADDRINUSE; that is the UID-wide name mutex now, instead of a side effect of a directory existing.
- **Pids come from the kernel, not from the supervisor.** `listen()` stamps the caller's pid into a socket and SO_PEERCRED hands it to whoever connects. The supervisor listens on the control socket; the child listens on the pgrp socket right after `setpgid()` and closes its copy. So `sandlock kill` learns the supervisor's pid and the child's process group from two `connect()` calls and never needs the supervisor to answer, which makes it work on a supervisor that is stopped or wedged.

## Design

- Every mode binds: supervised, `--no-supervisor`, and nested. Handlers that need supervisor state take `Option<Arc<SupervisorCtx>>`; without one, `config` serves the static policy and `ports` an empty map.
- Both sockets are bound before the fork, so a name collision fails with no child to reap. The child inherits the pgrp socket, calls `listen()` on it after `setpgid()`, then closes it; the supervisor keeps the fd for the life of the sandbox and drains the (never-spoken-on) connections it receives so the backlog stays empty.
- New `info` verb returns `{mode}`, replacing the mode marker. Wire protocol otherwise unchanged, still version 1.
- Abstract names have no permission bits and no owner, so authentication is mutual: the supervisor checks SO_PEERCRED on every accept and closes any connection from another uid without a response, and the client checks the listener's SO_PEERCRED before trusting either socket, so another uid cannot squat a name and feed `kill` pids of its choosing. Each request is bounded to five seconds so a stalled client cannot wedge introspection.
- `ps` reads `/proc/net/unix`, keeps listening sockets under `@sandlock/<uid>/` (skipping the `/pgrp` entries), takes the child pid from the pgrp socket, and asks the control socket for the mode and ports. A supervisor that does not answer within the timeout is still listed, as `unresponsive` with its real pid, since its socket proves it is alive.
- `kill` does `killpg(child)` and `kill(supervisor)` with the two peer pids and sends no request.
- `wait()` awaits the aborted control task so the names are released before it returns; running the same name twice in a row works. A forked child closes any inherited sandlock control socket other than its own pgrp socket as its first act, so a created-but-not-started sandbox (or a COW clone, which never execs) cannot pin a sibling's name, and a confined child never holds a handle to another supervisor's socket.
- Sandbox names additionally reject whitespace and control characters, because `/proc/net/unix` is whitespace-delimited.

Removed: the runtime dir helpers, `setup_runtime_dir`, `cleanup_runtime_dir`, `list_live_sandboxes`, `SandboxRuntime.control_dir`, and all prune logic. Pre-1.0 hard break.

## Tests

- Control suite rewritten around the new mechanism: list, inspect, ports, ps ports column, name collision in both modes, invalid names (now including whitespace), no_supervisor answering `info`, `config`, and an empty `ports`, with the supervisor pid checked against the spawned CLI process and the child pid checked to lead its own group.
- New: a SIGKILLed supervisor's name disappears and is reusable at once; a SIGSTOPped supervisor is listed as `unresponsive`; `sandlock kill` terminates a SIGSTOPped supervisor and its child and frees the name; the same name runs twice in a row within one process.
- Unit: `/proc/net/unix` parser against captured lines, including a `/pgrp` entry; bind is the name mutex and the listing follows the listener; the longest name plus the `/pgrp` suffix fits `sun_path`; the client reads both pids from the kernel with nothing serving the sockets, reports `still starting` before the child has listened, and refuses a listener owned by another uid; the server answers its own uid and closes on another without a byte; listing survives a foreign non-UTF-8 abstract name. The uid each side expects is a parameter, so the refusal paths run unprivileged on every CI job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,9 +771,9 @@ stop them:
 
 ```
 $ sandlock ps
-NAME                                  PID        UPTIME  CMD
-api.local                           12345            5m  python3 server.py
-web.local                           12346            3m  python3 server.py
+NAME                                  PID        UPTIME  STATUS      PORTS                     CMD
+api.local                           12345            5m  running     8080→41235                python3 server.py
+web.local                           12346            3m  running     8080→41236                python3 server.py
 
 $ sandlock inspect api.local --toml | head -10
 [config]
@@ -783,7 +783,7 @@ http_inject_ca = []
 ...
 
 $ sandlock kill web.local
-Killed sandbox 'web.local' (PID 12346)
+Killed sandbox 'web.local' (child PID 12346, supervisor PID 12340)
 ```
 
 This enables external reverse proxies (nginx, envoy) to route traffic
--- a/crates/sandlock-cli/src/main.rs
+++ b/crates/sandlock-cli/src/main.rs
@@ -974,6 +974,9 @@ fn validate_cli_name(name: &str) -> Result<(), String> {
     if name.contains('\0') {
         return Err("sandbox name must not contain NUL bytes".into());
     }
+    if name.bytes().any(|b| b <= 0x20 || b == 0x7f) {
+        return Err("sandbox name must not contain whitespace or control characters".into());
+    }
     if name.contains('/') {
         return Err("sandbox name must not contain '/'".into());
     }
--- a/crates/sandlock-cli/src/main.rs
+++ b/crates/sandlock-cli/src/main.rs
@@ -275,30 +275,40 @@ async fn main() -> Result<()> {
         }
 
         Command::Ps => {
-            match sandlock_core::control::list_live_sandboxes() {
-                Ok(sandboxes) if sandboxes.is_empty() => {
-                    println!("No running sandboxes.");
-                }
-                Ok(sandboxes) => {
-                    println!(
-                        "{:<32} {:>8}  {:>12}  {:<10}  {:<24}  {}",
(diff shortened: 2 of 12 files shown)


congwang-mk:

sachin2605:
Hi @congwang-mk  . This is a much more clean implemention for supervisor free mode. GTG from me.

PR #209, opened 2026-09-06
fix(learn): agent workload bug fixes

ghazariann:
Running a full learn-then-run cycle on `claude -p "hello"` exposed two bugs in `sandlock learn`.

### Bug 1 - `write = ["/"]`

The generated profile granted write access to the entire filesystem:

```toml
write = ["/"]
```

Two root causes:

1. `mkdirat("/root")` fires pre-syscall during learn. The target already exists (EEXIST), but the handler still inserted the parent `/` into the write set. Fix: skip `mkdirat`/`mknodat` if the target already exists.

2. The existing-path fast-path in `collapse_write_paths` had no check at all - only the ancestor-walking branch had guards. A direct write to `/` would pass straight through. Fix: check for `/` explicitly in the direct-write branch and drop it with a warning. Protected/Guarded paths in the direct-write branch emit a notice but are still recorded.

### Bug 2 - `/proc/self/maps` missing from reads

`canonicalize_or_keep` resolved `/proc/self/maps` to `/proc/<pid>/maps`, which `is_junk_path` then dropped as a pid-specific path. V8 reads `/proc/self/maps` on every startup - without it in the profile, Landlock denies the read and the process crashes.

Fix: don't resolve `/proc/self` paths; keep them as-is. Only junk volatile sub-trees (`/proc/self/fd/`, `/proc/self/task/`, etc.) and numeric-pid entries.

### Repro

```bash
# Learn
./target/debug/sandlock learn --http-inject-ca /etc/ssl/certs/ca-certificates.crt -o profile.toml -- claude -p "hello"

# Run
./target/debug/sandlock run --profile-file profile.toml -- claude -p "hello"
```

### DIFF
--- a/crates/sandlock-cli/src/learn.rs
+++ b/crates/sandlock-cli/src/learn.rs
@@ -378,7 +378,7 @@ impl LearnObserver {
             }
             "openat" | "open" => {
                 if let Some(path) = event.path {
-                    let path = canonicalize_or_keep(path);
+                    let path = canonicalize_or_keep(path, event.pid);
                     if let Some(fl) = event.flags {
                         if is_write_open(fl) {
                             self.writes.lock().unwrap().insert(path);
@@ -392,7 +392,7 @@ impl LearnObserver {
             // rights, so the parent dir is what sandlock run needs, not the target.
             "mkdirat" | "mknodat" => {
                 if let Some(p) = event.path {
-                    let p = canonicalize_or_keep(p);
+                    let p = canonicalize_or_keep(p, event.pid);
                     // If the target already exists the syscall will fail with
                     // EEXIST; no MAKE_DIR right on the parent is needed.
                     if !p.exists() {
@@ -426,7 +426,7 @@ impl LearnObserver {
             // dst operates on the link itself, so only the parent is canonicalized.
             "linkat" => {
                 if let Some(src) = event.path {
-                    self.reads.lock().unwrap().insert(canonicalize_or_keep(src));
+                    self.reads.lock().unwrap().insert(canonicalize_or_keep(src, event.pid));
                 }
                 if let Some(dst) = event.path2 {
                     let dst = canonicalize_parent_or_keep(dst);
@@ -438,7 +438,7 @@ impl LearnObserver {
             // truncate: LANDLOCK_ACCESS_FS_TRUNCATE applies to the file itself.
             "truncate" => {
                 if let Some(p) = event.path {
-                    self.writes.lock().unwrap().insert(canonicalize_or_keep(p));
+                    self.writes.lock().unwrap().insert(canonicalize_or_keep(p, event.pid));
                 }
             }
             "bind" => {
@@ -449,7 +449,7 @@ impl LearnObserver {
                 } else if let Some(p) = event.path {
                     // AF_UNIX named bind: Landlock MAKE_SOCK is a directory right,
                     // so the parent dir is what sandlock run needs.
-                    let p = canonicalize_or_keep(p);
+                    let p = canonicalize_or_keep(p, event.pid);
                     if let Some(parent) = p.parent() {
                         self.writes.lock().unwrap().insert(parent.to_path_buf());
                     }
@@ -515,12 +515,22 @@ impl LearnObserver {
 /// Resolve symlinks to get the canonical path. Falls back to the original
 /// if the path doesn't exist yet (e.g. COW-intercepted creates).
 ///
-/// /proc/self is deliberately not resolved: canonicalize would turn
-/// /proc/self/maps into /proc/<pid>/maps, making it look like a volatile
-/// pid-specific path and causing is_junk_path to drop it.
-fn canonicalize_or_keep(p: PathBuf) -> PathBuf {
-    if p.as_os_str().as_encoded_bytes().starts_with(b"/proc/self") {
-        return p;
+/// /proc/self paths are canonicalized via the event pid so symlinks
+/// (e.g. /proc/self/exe) resolve against the workload, not the supervisor.
+/// Results still under /proc/<pid>/ are mapped back to /proc/self/.
+fn canonicalize_or_keep(p: PathBuf, pid: u32) -> PathBuf {
+    let b = p.as_os_str().as_encoded_bytes();
+    if b.starts_with(b"/proc/self") {
+        let suffix = &b[b"/proc/self".len()..];
+        let pid_path = PathBuf::from(format!("/proc/{}{}", pid, String::from_utf8_lossy(suffix)));
+        let resolved = std::fs::canonicalize(&pid_path).unwrap_or(pid_path);
+        let resolved_b = resolved.as_os_str().as_encoded_bytes();
+        let pid_prefix = format!("/proc/{}/", pid);
+        if resolved_b.starts_with(pid_prefix.as_bytes()) {
+            let rest = &resolved_b[pid_prefix.len() - 1..]; // keep leading /
+            return PathBuf::from(format!("/proc/self{}", String::from_utf8_lossy(rest)));
+        }
+        return resolved;
     }
     std::fs::canonicalize(&p).unwrap_or(p)
 }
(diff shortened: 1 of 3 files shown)


docs/learn.md
```diff
@@ -53,12 +53,17 @@ optional, so omitting the ancestor would cause `sandlock run` to abort.
...
-| Tier | Paths | Write (auto) | `--collapse N` | `--collapse-prefix` |
+| Tier | Paths | Write collapse | `--collapse N` | `--collapse-prefix` |
 |---|---|---|---|---|
 | **Protected** | `/`, `/root`, `~/.ssh`, `~/.aws`, `~/.kube`, `~/.gnupg` | skip + error | never (keep individual file) | refused unless `--force-sensitive-collapse` |
 | **Guarded** | `/etc`, `/proc`, `/sys`, `/dev`, `/boot`, `/run/secrets` | emit + warning + diff | never (keep individual file) | refused unless `--force-sensitive-collapse` |
 | **Normal** | everything else | collapse freely | collapse freely | collapse freely |
 
+The tiers apply to **write collapse only** (when a non-existent path's nearest
+existing ancestor is used as the Landlock grant). Direct writes to an existing
+path — including Protected and Guarded paths — are recorded as-is without
```
congwang-mk: This contradicts the code in this PR: the existing-path branch of collapse_write_paths now skips Protected and warns on Guarded, so direct writes are filtered. Please rewrite. Also drop the em-dashes; the project does not use them.

crates/sandlock-cli/src/learn.rs
```diff
@@ -75,6 +75,23 @@ fn collapse_write_paths(writes: &BTreeSet<PathBuf>) -> Vec<PathBuf> {
     for p in writes {
         if is_junk_path(p) { continue; }
         if p.exists() {
+            match classify_path(p) {
+                PathTier::Protected => {
+                    eprintln!(
+                        "sandlock learn: WARNING: observed a write of '{}' (protected path), skipping",
```
congwang-mk: Per the PR body, /root landed in the write set because the workload itself renamed into it. Dropping it makes the generated profile unable to satisfy that rename, so sandlock run will fail on it, yet learn still exits 0. The docs table promises "skip + error" for this tier. Either exit non-zero or make the warning say the profile is known incomplete. What was the rename target in your repro, given you report the full cycle works?

crates/sandlock-cli/src/learn.rs
```diff
@@ -189,13 +206,24 @@ fn collapse_by_threshold(
...
 fn is_junk_path(p: &std::path::Path) -> bool {
     let b = p.as_os_str().as_encoded_bytes();
-    // /proc/self/... and /proc/<pid>/... are pid-specific;
-    let proc_pid = b.starts_with(b"/proc/self")
-        || (b.starts_with(b"/proc/") && b.get(6).map_or(false, u8::is_ascii_digit));
+    // /proc/<pid>/... are pid-specific across runs.
+    let proc_numeric_pid = b.starts_with(b"/proc/")
+        && b.get(6).map_or(false, u8::is_ascii_digit);
+    // Under /proc/self, only sub-trees with volatile numeric components are
+    // junk. Stable entries like /proc/self/maps legitimately needed across runs (e.g. V8 reads maps on every startup).
```
congwang-mk: Does this grant cover child processes? Landlock is applied in the forked child before exec, so a rule on /proc/self/maps binds to that one pid's proc inode. Any subprocess the workload spawns has a different /proc/<pid>/maps inode and would still be denied. Worth checking with a workload that forks a node child.

crates/sandlock-cli/src/learn.rs
```diff
@@ -483,7 +515,14 @@ impl LearnObserver {
 
 /// Resolve symlinks to get the canonical path. Falls back to the original
 /// if the path doesn't exist yet (e.g. COW-intercepted creates).
+///
+/// /proc/self is deliberately not resolved: canonicalize would turn
+/// /proc/self/maps into /proc/<pid>/maps, making it look like a volatile
+/// pid-specific path and causing is_junk_path to drop it.
 fn canonicalize_or_keep(p: PathBuf) -> PathBuf {
+    if p.as_os_str().as_encoded_bytes().starts_with(b"/proc/self") {
```
congwang-mk: Minor: this also leaves /proc/self/cwd/x and /proc/self/exe unresolved. Not a regression, since canonicalize previously resolved these against the supervisor's own /proc/self, which was the wrong process anyway. A cleaner fix would rewrite the prefix to /proc/<event.pid>/, canonicalize, then map back to /proc/self.

congwang-mk (changes requested):
The two fixes look right and CI is green, but a few things to address before merge. Also: none of the three behavior changes here (existing-path tier check, mkdirat on an existing target, /proc/self surviving is_junk_path) has a test; learn_test.rs already covers the ancestor-walk branch, so adding cases next to those would be natural. The three commits also move /root guarded, then protected, and end with classify_path identical to main; please squash into logical commits with a body explaining why.

crates/sandlock-cli/src/learn.rs
ghazariann: You are right. I tested with a simple program forking a child that reads `/proc/self/maps`: parent ok, child fails with `Permission denied`. The Landlock grant pins the top-level process's `/proc/<pid>/maps` inode; children get different inodes not covered by the rule.

One option during learn: when a `/proc/self/` access is observed from a child process, widen the grant to /proc. Simple but broad. Another option: intercept these opens at run time via seccomp notif and handle them in the supervisor instead. Any thoughts ?

congwang-mk:
@ghazariann Thanks for the update, please also do a rebase since there is a conflict.

crates/sandlock-cli/src/learn.rs
ghazariann: Done e13cd14e8

crates/sandlock-cli/src/learn.rs
ghazariann: I initially thought the profile was complete because `sandlock run` exited 0 and claude responded normally even without `/root` in the write list, but the rename of `.claude.json.tmp` → `.claude.json` was silently failing. Claude somehow swallowed the `EACCES` and continued. I confirmed this by checking `.claude.json`'s mtime before and after the run: it didn't change.

After looking at this more closely, I removed all filtering from the direct-write branch (4f8f74d7706ebd6), only `/` is still dropped, since granting it subsumes everything else. After all, `sandlock learn` is an observer: if a legitimate workload wrote to `/root`, the user should see it and decide. Thoughts?

docs/learn.md
ghazariann: Now the doc is in sync with the PR.

crates/sandlock-cli/src/learn.rs
congwang-mk: Thanks for confirming. Please go with the seccomp option, not the /proc widening: in non-chroot mode a /proc read grant exposes every same-uid process's /proc/<pid>/ to the workload, which is exactly what the Guarded tier exists to flag.

The seccomp route is smaller than it sounds because the hook already exists. `procfs::handle_proc_open` runs on every open-family syscall in every mode and already injects memfds for virtualized files. The extension would be: when the profile's read list contains `/proc/self/X` and the calling process opens `/proc/self/X`, `/proc/thread-self/X`, or `/proc/<its own pid>/X`, the supervisor opens `/proc/<notif.pid>/X` and injects that fd. That is scoped to the caller's own pid and to paths the policy explicitly lists, so it does not reopen the issue #27 concern about on-behalf opens bypassing Landlock.

That belongs in a follow-up PR: it changes what `run` enforces rather than what `learn` records, lives in a different crate, and needs its own test with a forking workload. For this PR, please add a one-line caveat to docs/learn.md that a `/proc/self/*` grant currently only covers the top-level process, or open a tracking issue and link it here.

crates/sandlock-cli/src/learn.rs
congwang-mk: Agreed, this is the right call. My objection was that dropping `/root` produced an incomplete profile while learn still exited 0. Recording it removes the incompleteness and the NOTE gives the operator the visibility, which is better than either option I suggested.

One thing to keep in mind for later, not for this PR: `mkdir /root/foo` now grants `/root` with a NOTE, while `open("/root/foo/bar", O_CREAT)` with `/root/foo` missing goes through the ancestor walk and hits the Protected skip + error. Same effective grant, two outcomes. That inconsistency predates this PR, so a tracking issue is enough.

ghazariann: Tracking #217

crates/sandlock-cli/src/learn.rs
ghazariann: Tracking #218

PR #220, opened 2026-09-10
learn: follow-ups from the #209 review

congwang-mk:
Follow-ups to #209 that were left out of that PR as minor.

- `canonicalize_or_keep` rewrote `/proc/self` to `/proc/<pid>` but only mapped results back when they carried a trailing `/<pid>/` prefix, so an open of `/proc/self` itself came out as a numeric-pid path and was dropped as junk. Both sides now use `Path::strip_prefix`, so the bare directory and its entries round-trip the same way, and unrelated names that merely start with `/proc/self` no longer match.
- `canonicalize_parent_or_keep` still resolved against the supervisor's own `/proc/self`; it now goes through the same pid-aware helper, so rename and unlink targets under `/proc/self/cwd` land on the workload's process.
- The `/proc/self/exe` test only asserted that the alias was absent, which a dropped read would also satisfy. It now requires the resolved interpreter path.
- The three `/` guards in learn shared one rule spelled two ways; they now call a single `is_fs_root` helper. This only affects `sandlock learn` output, not `run` enforcement.
- `docs/learn.md` said a warning is printed for direct writes to Protected or Guarded paths; the code prints a NOTE and still records the path.

Tests: `cargo test -p sandlock-cli --test learn_test -- --test-threads=4` passes (40/40), including a new `test_proc_self_bare_preserved` that fails on main.

### DIFF
--- a/crates/sandlock-cli/src/learn.rs
+++ b/crates/sandlock-cli/src/learn.rs
@@ -4,7 +4,7 @@
 //! usable by `sandlock run -p`.
 
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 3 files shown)


ghazariann:
Lgtm

PR #228, opened 2026-09-14
network: allow combining net allow and deny policies

### LINKED ISSUE #135: Allow to combine --net-allow and --net-deny
Requested by @SuperManito. Currently they are exclusive, because --net-allow means defaulting to deny, --net-deny mean defaulting to allow. Although combining them could be powerful, the logic might be too complicate if we don't design it right. Also, the implementation would be complicate too. We need to evaluate this carefully and maybe come up with a better design.

bferanmi806-sketch:
Fixes #135.

Outbound policy is now allow-only, deny-only, or combined allow-minus-deny: with both lists present a destination must match an allow rule and must not match a deny rule. The same precedence applies to TCP bind allow/deny, where denied ports win.

Architecture (per congwang-mk review 2026-09-18): Sandbox.net_allow holds only explicit user rules. HTTP reachability is generated at resolution time via http_net_allow_rules() from http_allow/http_deny/http_ports and merged only at consumption sites (sandbox spawn resolution including virtual /etc/hosts, Landlock connect-port gates, resolved feature flags). The mode is derived from the separated configuration with no net_allow_explicit flag, no custom trailing deserializer, no checkpoint/meta.json persistence or backfill, and no profile/CLI special cases: explicit allow always activates the layer; HTTP-derived reachability activates it only when no deny is present (HTTP-only stays a restrictive allowlist, deny-only+HTTP stays default-allow).

Checkpoint: IMAGE_VERSION bumped 2 -> 3. Upstream v2 images may store HTTP-derived rules inside net_allow, which the separated representation would misread as explicit rules (deny-only+HTTP loading as combined), so v2 is rejected outright with no compatibility fallback (pre-1.0 policy). Regression test: image_rejects_v2_http_mixed_net_allow.

Static deny stays effective after dynamic allow resolution: the effective allow resolves per-PID override > live policy > static allowlist, then the immutable static deny is checked and wins, including under Unrestricted allow.

Combined finite TCP policies keep the Landlock allow-side port gate; deny-only and all-ports/wildcard allow drop CONNECT_TCP so the on-behalf path enforces.

Regression coverage: combined finite overlap/non-overlap, wildcard and Unrestricted allow with deny carve-outs, hostname resolving into a denied CIDR, deny-only with HTTP, HTTP-only, combined+HTTP, profile/CLI rebuild round-trips, checkpoint round-trips plus v2 rejection, and Landlock mode selection (deny-only, combined finite/wildcard, HTTP-only, deny-only+HTTP, combined+HTTP) including ephemeral bind(0).

Validation (WSL Ubuntu, Rust 1.98.1): cargo test -p sandlock-core --lib => 802 passed, 0 failed. Sandbox-execution CLI/integration tests cannot run in this WSL host (Landlock ABI v3, NetTcp requires v4); expected in CI. Go/Python sources untouched by this change.


congwang-mk:
@bferanmi806-sketch Thanks for the PR! Please fix the failing tests and rebase to solve the conflicts. I will look into it tomorrow.

bferanmi806-sketch:
Thanks! I’ve rebased this onto the latest `main` and resolved the conflicts with the recent bind enforcement changes.

The learn-mode bind regression is covered by preserving the upstream wildcard bind allowance, and I added regression coverage around the combined bind policy and the `NetTcp` protection guard as well.

Local build/core validation is passing as far as this host supports. The new Actions run is currently waiting for approval to start on the fork, so there aren’t any CI jobs/results yet.

Should be ready for another look when you get a chance.

congwang-mk:
Thanks for the rebase. Before a line-level pass I want to settle two design points.

**1. The `net_allow_explicit` flag**

This flag exists only because `extend_net_allow_for_http` appends generated reachability rules into the user's `net_allow`, so the code can no longer tell "the user gave allow rules" from "HTTP generated them". The PR patches that by recording an origin bit and threading it through the builder, `Clone`, the checkpoint blob, `meta.json`, profile output, the CLI profile re-feed, `resolved.rs`, and `landlock.rs`, with a tri-state whose `None` means different things depending on whether `net_deny` is present.

Please fix this at the root instead: keep HTTP-derived rules out of `net_allow` (a separate field, or generate them at resolve time from `http_allow`/`http_deny`/`http_ports`) and merge them only where the rules are consumed (`resolve_net_allow` in `sandbox.rs` and the feature flags in `resolved.rs`). Then the mode is derivable from the two lists alone, and the flag, the custom deserializer, and every special case in `profile.rs` and `main.rs` go away.

**2. The compatibility machinery**

Sandlock does not keep backward compatibility before 1.0, so please drop all of it:

- `deserialize_trailing_net_allow_explicit` decides "legacy blob" by string-matching serde error text (`"eof"`, `"failed to fill"`, ...). That is fragile and not something we want in the policy loader.
- The `meta.json` backfill is dead code: a legacy image has neither the trailing blob field nor the meta key, so there is nothing to backfill from.
- If a new field must land in the checkpoint policy, bump `IMAGE_VERSION` and let old images be rejected; the loader already does that.

With point 1 done, none of this is needed anyway.

I will also approve the CI run once these are addressed, and then review the rest.

bferanmi806-sketch:
Thanks for the detailed feedback! I've addressed both design points.

HTTP-derived reachability is now generated separately at resolution time, keeping `Sandbox.net_allow` explicit-only. I've also removed the `net_allow_explicit` flag and the associated compatibility machinery.

I bumped `IMAGE_VERSION` to 3 to prevent older checkpoints from being misinterpreted under the new policy semantics.

The Rust library tests pass locally (802 passed, 0 failed). The Linux integration tests still need CI validation, as my WSL host only supports Landlock ABI v3.

The changes are pushed, and I've updated the PR description accordingly. Could you approve the pending CI run when you get a chance?

Also, if there are any other issues or areas of Sandlock where you could use an extra pair of hands, I'd be happy to help. I'm keen to keep contributing to the project!

Thanks again!

congwang-mk (commented):
Thanks, both design points are addressed and the separation looks right. Remaining items are inline. One that has no line to anchor to:

**Commit history.** Please squash into 2 or 3 logical commits. Right now there are fixup commits, one with the bare subject `docs:`, early commits that add `net_allow_explicit` only for later ones to remove it, and the body of 884356f says "IMAGE_VERSION stays 2" which the next commit contradicts. Every commit also needs a `Signed-off-by` (`git commit -s`).

crates/sandlock-core/tests/integration/test_network.rs
```diff
@@ -67,8 +67,8 @@ async fn test_udp_rule_scopes_destination_by_host() {
 /// kernel's `addrlen > sizeof(sockaddr_storage) -> EINVAL` check, so without the
 /// `read_sockaddr` guard the supervisor did `vec![0u8; 0xFFFFFFFF]` (~4 GiB) and
 /// could OOM/abort, taking down every sandbox. The child sends with a bogus 4 GiB
-/// `addr_len` via raw `libc.sendto`. The supervisor must (a) survive — the whole
-/// run completes — and (b) reject the oversized length with `EINVAL` (22),
+/// `addr_len` via raw `libc.sendto`. The supervisor must (a) survive â€” the whole
```
congwang-mk: Encoding damage: 24 comment lines in this file that the PR does not otherwise touch had their dashes turned into mojibake (looks like a Windows editor re-save). Please revert all of those hunks.

crates/sandlock-core/src/policy_fn.rs
```diff
@@ -481,7 +481,12 @@ mod tests {
         let ceiling = test_live();
         let pid_overrides = Arc::new(RwLock::new(HashMap::new()));
         let denied = Arc::new(crate::seccomp::state::DeniedSet::default());
-        let mut ctx = PolicyContext::new(live.clone(), ceiling, pid_overrides, denied);
+        let mut ctx = PolicyContext::new(
```
congwang-mk: Unrelated churn: these six hunks only re-wrap `PolicyContext::new(...)`, left over from a removed parameter. Please revert, along with the same re-wrap of `spawn_policy_fn` in `sandbox.rs`, the `let nr` move in `seccomp/notif.rs`, and the docstring deletions in `test_policy_fn.rs` and `python/tests/test_policy_fn.py`.

crates/sandlock-core/src/seccomp/dispatch.rs
```diff
@@ -729,12 +729,12 @@ pub(crate) fn build_dispatch_table(
...
-            let sup = Arc::clone(&__sup);
-            let notif_fd = cx.notif_fd;
-            async move {
-                crate::port_remap::handle_bind(&notif, &sup.network, notif_fd).await
-            }
-        });
+                let sup = Arc::clone(&__sup);
+                let notif_fd = cx.notif_fd;
+                async move {
+                    crate::port_remap::handle_bind(&notif, &sup, notif_fd).await
```
congwang-mk: This block is now indented one level too deep. Only the `&sup.network` to `&sup` argument change is needed here.

crates/sandlock-core/src/port_remap.rs
```diff
@@ -140,63 +142,82 @@ pub(crate) async fn handle_bind(
...
+        let live_policy = {
+            let pfs = ctx.policy_fn.lock().await;
+            pfs.live_policy.clone()
+        };
+        let udp_deny_all = protocol == Some(crate::network::Protocol::Udp)
+            && ns
+                .effective_network_policy(
+                    notif.pid,
+                    crate::network::Protocol::Udp,
+                    live_policy.as_ref(),
```
congwang-mk: Behavior change unrelated to #135: main passes `None` here. With a non-empty policy_fn live allowlist the UDP layer becomes an IP-only allow, `denies_everything()` turns false, and UDP binds start succeeding under TCP-only rule sets. Please revert to `None`, or split it out with a test and a justification.

crates/sandlock-core/src/seccomp/state.rs
```diff
@@ -489,35 +518,48 @@ impl NetworkState {
...
+        let deny = match protocol {
+            Protocol::Tcp => self.tcp_deny_policy.clone(),
+            Protocol::Udp => self.udp_deny_policy.clone(),
+            Protocol::Icmp => self.icmp_deny_policy.clone(),
+        };
         let ip_only_allow = |ips: &HashSet<std::net::IpAddr>| {
-            let per_ip = ips.iter().map(|&ip| (ip, PortAllow::Any)).collect();
+            let per_ip = ips
+                .iter()
+                .map(|ip| (ip.to_canonical(), PortAllow::Any))
```
congwang-mk: `to_canonical()` is also unrelated to this PR and untested. Please drop it or send it separately with a test.

crates/sandlock-core/src/checkpoint/image.rs
```diff
@@ -310,4 +338,100 @@ mod tests {
         let msg = res.unwrap_err().to_string();
         assert!(msg.contains("version"), "error should mention version, got: {msg}");
     }
+
+    #[test]
+    fn image_rejects_v2_http_mixed_net_allow() {
```
congwang-mk: This duplicates `image_rejects_wrong_version`: the fixture is just a `meta.json` with `version: 2` and contains no `net_allow`, so the name overpromises. Please drop it.

crates/sandlock-core/src/checkpoint/image.rs
```diff
@@ -310,4 +338,100 @@ mod tests {
...
+            msg.contains("version") && msg.contains('2'),
+            "v2 image must be rejected as unsupported version, got: {msg}"
+        );
+    }
+
+    fn round_trip_policy(policy: Sandbox) -> Sandbox {
+        let dir = std::env::temp_dir().join(format!(
+            "sandlock-mode-{}-{}",
+            std::process::id(),
+            // Nanos Erbatur to avoid collisions when tests run in parallel.
```
congwang-mk: Garbage comment ("Nanos Erbatur"). Please remove.

crates/sandlock-core/src/sandbox.rs
```diff
@@ -2097,13 +2179,13 @@ impl Sandbox {
...
-                net_state.icmp_policy = resolved_deny.icmp;
-            } else {
-                let no_rules = self.net_allow.is_empty();
+            let resolved_deny = network::resolve_net_deny(&self.net_deny);
+            net_state.tcp_deny_policy = resolved_deny.tcp;
+            net_state.udp_deny_policy = resolved_deny.udp;
+            net_state.icmp_deny_policy = resolved_deny.icmp;
+
+            if self.net_allow_is_active() {
+                let no_rules = effective_net_allow.is_empty();
```
congwang-mk: Dead code: inside `if self.net_allow_is_active()` the effective list can never be empty, so `no_rules` is always false. Drop it and test only `resolved.any_ip_all_ports`.

python/README.md
```diff
@@ -576,13 +578,17 @@ Methods available inside `policy_fn`:
...
-| `ctx.grant_network(ips)` | Allow additional IP addresses |
+| `ctx.grant_network(ips)` | Allow additional IP addresses up to the static network ceiling |
 | `ctx.restrict_max_memory(bytes)` | Lower memory limit |
 | `ctx.restrict_max_processes(n)` | Lower process limit |
 | `ctx.restrict_pid_network(pid, ips)` | Per-PID network restriction |
 | `ctx.deny_path(path)` | Deny access to a path |
 | `ctx.allow_path(path)` | Remove a previously denied path |
 
+Dynamic IP restrictions resolve with legacy priority (per-PID override > live
+policy > static allowlist); the static `net_deny` layer is always checked
```
congwang-mk: Says the static deny is "always checked first", but the code and the other docs say it runs after the allow verdict (a852a06 meant to fix exactly this wording).

docs/sandbox-reference.md
```diff
@@ -533,7 +536,14 @@ parse_ports([80, "443", "8000-8005"])
...
    of that protocol is configured.
-2. **Seccomp COW with `workdir`.** When `workdir` is set, the
+2. **Allow/deny precedence.** `net_allow` and `net_deny` may be used
+   together; a destination must match the allowlist and must not match the
+   denylist. The same precedence applies to `net_allow_bind` and
+   `net_deny_bind` for TCP ports. `policy_fn` IP restrictions resolve with
+    legacy priority (per-PID override > live policy > static allowlist) and
+    the static denylist is applied after the allow verdict and always wins,
+    so a dynamic override can never erase it.
+3. **Seccomp COW with `workdir`.** When `workdir` is set, the
```
congwang-mk, line 546: The list now has two items numbered 3 (this one and "HTTP host auto-expansion"), and the continuation lines of the new item 2 are indented 4 spaces instead of 3.

go/README.md
```diff
@@ -106,11 +106,12 @@ policy on each call.
 host, IP, or CIDR (`"10.0.0.0/8:443"`, `"[2606:4700::/32]:443"`); scheme
 prefixes opt other protocols in (`"udp://1.1.1.1:53"`, `"udp://*"`,
 `"icmp://host"`, `"icmp://*"`). `NetDeny` is the inverse (default-allow
-denylist, IP/CIDR targets only, mutually exclusive with `NetAllow`).
+ denylist, IP/CIDR targets only); when both are set, denied destinations win.
```
congwang-mk: Stray leading space here and on the two `NetDenyBind` lines below.

crates/sandlock-cli/src/main.rs
```diff
@@ -491,6 +491,8 @@ async fn run_command(args: RunArgs) -> Result<i32> {
         for p in &base.fs_readable { b = b.fs_read(p); }
         for p in &base.fs_writable { b = b.fs_write(p); }
         for p in &base.fs_denied { b = b.fs_deny(p); }
+        // `Sandbox.net_allow` holds only explicit rules; HTTP reachability is
```
congwang-mk: Nit: this "net_allow holds only explicit rules" note is repeated in `main.rs`, `profile.rs`, `builder.rs`, `http.rs`, `image.rs` and twice in `sandbox.rs`. The field doc on `Sandbox.net_allow` is enough; please remove the rest.

crates/sandlock-core/src/sandbox.rs
```diff
@@ -672,6 +695,62 @@ impl Sandbox {
...
+        out
+    }
+
+    /// Whether the outbound allow rules are an active runtime layer.
+    ///
+    /// Derived from the separated configuration: explicit `--net-allow` rules
+    /// always activate the layer; HTTP-derived reachability activates it only
+    /// for non-deny policies (HTTP-only stays a restrictive allowlist, while
+    /// deny-only+HTTP stays default-allow with the HTTP proxy on top).
+    pub fn net_allow_is_active(&self) -> bool {
```
congwang-mk: Nit: this is `pub` only so the integration tests can assert on it. Prefer `pub(crate)` and let those tests assert through behavior (they already do), or mark it `#[doc(hidden)]`.

crates/sandlock-core/src/port_remap.rs
```diff
@@ -140,63 +142,82 @@ pub(crate) async fn handle_bind(
...
-    // under an allowlist because Landlock refuses it too: only `'*'` can
-    // express "any ephemeral port".
-    let denied = match ip_port {
-        Some(port) => {
-            let ns = network.lock().await;
-            ns.bind_allow_ports.as_ref().is_some_and(|allow| !allow.contains(&port))
-                || ns.bind_deny_ports.contains(&port)
-        }
-        None => false,
+    // Non-IP families: bind verbatim — there is no TCP port policy to apply.
```
congwang-mk: Nit: please avoid em dashes in new comments (here and at line 202); use a colon or a comma.

bferanmi806-sketch:
Thx i addedd the fix if there are any road map items or issues u need help with feel free to assign them to me

